PR #13, opened 2026-06-06
fix(network): roll schema_version back to 8 (unblock deployed fleet)

jhumel-code:
## Why

The `schema_version: 9` bump (commit 81e5a56, merged in #11) shipped to rules `main` **before** an engine release that implements the `has_http_call_without_timeout` predicate it introduced. The deployed engine — **v0.1.2, `SupportedSchemaVersion = 8`** — rejects any pack whose `schema_version` exceeds 8, so it discards the *entire* pack as incompatible.

Consequence: the **default scan path is live-broken**. `trustabl scan` (and the GitHub Action, which runs v0.1.2 on a fresh runner with no cache and pulls rules `main`) fail with *"rules are newer than this version; upgrade Trustabl or pull an older `--rules-ref`."*

## What

`git revert` of 81e5a56:
- `manifest.yaml`: `schema_version` 9 → **8**
- `openai_sdk/network.yaml`: OAI-016 back to its v8 `has_body_text` form
- `vercel_ai/network.yaml`: VAI-011 dropped (v9-only predicate)

This restores the deployed fleet immediately, **no binary upgrade required** (consistent with the no-forced-upgrade principle). The v9 network rules return once the forward-compatible loader (Option C / TR-200) ships in the engine and can either evaluate the predicate or skip it gracefully.

Precedent: `fd790fa` previously rolled `schema_version` back to 8 for the same reason.


jhumel-code:
Closing: superseded by reality. The engine already ships schema-9 support + forward-compatible lenient loading in released v0.1.3 (SupportedSchemaVersion=9, has_http_call_without_timeout predicate, LoadLenient). rules main @ schema 9 loads cleanly on v0.1.3, and the GitHub Action's default version: latest resolves to v0.1.3 — so there is no live break. Rolling schema back to 8 would only strip the v9 network rules (VAI-011, OAI-016) the released engine can now evaluate. Consumers pinned to <= v0.1.2 should upgrade or pin --rules-ref instead.

PR #48, opened 2026-08-20
Ship compliance mapping packs and the OAI composition rules

jhumel-code:
Brings `feat/severity-reclassification-acac` up onto current `main`. Every item
here already existed on that branch but was unreachable from `main`, so no user
was receiving any of it.

## What lands

**Compliance mapping packs (new).** `mappings/*.yml` for NIST SP 800-53 r5,
ISO/IEC 27002:2022, SOC 2 (TSC 2017), the EU AI Act and PCI DSS. These are the
SME-reviewed crosswalk the engine's `compliance build-pack --mappings` path
consumes. `main` carried no `mappings/` directory at all, so that documented
path had no data behind it.

**Four rules that were tested but never shipped.** OAI-025 (dead tool via
call-graph liveness), OAI-107 (handoff target wires shell tools), OAI-112 and
OAI-113 (transitive shell through handoff chains). These are the only rules
anywhere exercising `reachability_is` and `transitive_capability_exceeds_direct`,
so both predicates currently reach no user.

**Severity reclassification.** Six permission-bypass / unrestricted-shell rules
move from high to critical: CSDK-103, CSDK-120, CSDK-201, CSDK-202, ADK-008 and
LC-101.

## Merge conflict resolution

One conflict, in ADK-008. `main` had corrected the title and widened the matcher
to cover both `ExecuteBashTool` (current google-adk) and `BashTool` (earlier
releases); the branch had only the older `BashTool`-only matcher plus the
severity bump. Resolved by keeping `main`'s corrected title, explanation and
wider matcher, and taking only the severity bump from the branch.

## Verification

`trustabl rules validate` output on this branch is byte-identical to its output
on clean `origin/main`, so this introduces no new schema, duplicate-ID or
unknown-predicate errors. The 12 pre-existing errors are the known
missing-predicate drift (CSDK-017/018, CSDK-204, CSKILL-080..086) and are
untouched here.

## Follow-up needed before this is fully compliant with the three-repo contract

The paired rulebook has not been updated yet:

- The six reclassified rules still carry `severity: high` in their rulebook
  front-matter, which `tools/check_rulebook.py` fails on.
- OAI-025, OAI-107, OAI-112 and OAI-113 have no rationale doc at all.

`schema_version` stays at 14, so the engine fixture (19) remains divergent.


jhumel-code:
Superseded by #49. This PR also carried the compliance crosswalk, which has moved to the engine repo: compliance is a Guard-only surface, so the mappings served no user here while being publicly exposed. #49 is the same rule changes without them.

PR #51, opened 2026-08-24
feat(google_adk): add ADK-111, MCPToolset with no tool_filter

ivanpaghubasan:
Closes Class 2 of the tool access-control detection scoping — Google ADK genuinely has a 'true restriction' semantic (unlike Claude/OpenAI's permission-style allow-lists), making 'no explicit filter' the correct risk signal.

MCPToolset connects an agent to a remote MCP server's entire tool catalog, which is not enumerable from source and can change whenever the server updates. tool_filter= is the only mechanism narrowing that catalog to a named allow-list.

Mirrored byte-for-byte from the engine fixture (verified via diff, no output — fully in sync).

### DIFF
--- a/google_adk/agent_safety.yaml
+++ b/google_adk/agent_safety.yaml
@@ -5,7 +5,8 @@ policy:
   description: >
     Agent-scope safety rules for Google ADK LlmAgents and workflow
     agents. Covers description hygiene, callback-based interception
-    of shell tools, and sub-agent shell exposure.
+    of shell tools, sub-agent shell exposure, and unrestricted MCP
+    tool surfaces.
 
 rules:
   - id: ADK-101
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @ivanpaghubasan. I reviewed this together with trustabl/trustabl#120, and the YAML here is byte-identical to the fixture copy in that PR, so the sync contract is satisfied by the pair. The rule itself is clean: every required field is present, `adk_llm_agent` is a valid agent-scope token, both predicates already exist in the engine, and severity high with confidence 0.75 is consistent with its neighbors in this pack. The explanation and fix read well and are framed for ADK specifically, naming the actual mechanism (`tool_filter=`) rather than a generic allow-list lecture. I also checked the open PR queue and ADK-111 is claimed only by this pair, so there is no ID collision to coordinate.

One ordering note: this needs to land together with or after trustabl/trustabl#120. Without the engine change, `MCPToolset` is never classified as a hosted tool, so the rule would ship but silently never fire, and the engine's rules-sync job would go red until the fixture catches up. I flagged two small doc updates on the engine side over there. The rule also needs a rationale doc in the rulebook before it ships; we can handle or pair on that part. Approving on the content, merging in step with the engine PR.

PR #52, opened 2026-08-24
feat(langchain): add LC-007, LC-015 tool network timeout rules

bradAGI:
LangChain is the only pack with tool discovery and no network-timeout rule. Pydantic AI (PYD-006), AutoGen (AG2-012), and the Vercel AI SDK (VAI-011) all ship one; this closes the gap for both LangChain languages.

A tool request with no deadline stalls the `AgentExecutor` iteration it runs in until the connection dies. LC-102's `max_iterations` cap doesn't help — it bounds how many steps the executor takes, not the wall-clock of one step — so the run hangs with nothing reported back to the model.

Follows the existing python/TypeScript pairing in this pack (LC-003/011, LC-004/012, LC-005/013): `call_without_kwarg` with the same callee list as PYD-006 for Python, `has_http_call_without_timeout` for LangChain.js.

**Verification** — built the engine at `main` and ran against fixtures:

```
$ trustabl rules validate .
OK: 86 rule pack(s), 208 rule(s) valid under rule schema version 14
```

Fire (`requests.get(url)` / bare `fetch(...)`): `LC-005, LC-007, LC-013, LC-015, LC-201`
Silent (`timeout=10` / `signal: AbortSignal.timeout(15_000)`): `LC-005, LC-013, LC-201`

No engine-side predicates needed — both are already shipped and used by other packs, so no `schema_version` bump.

### DIFF
--- a/langchain/network.yaml
+++ b/langchain/network.yaml
@@ -1,0 +1,83 @@
+policy:
+  id: langchain_network
+  name: LangChain tool network hygiene
+  category: langchain
+  description: >
+    Network-call hygiene inside LangChain tool functions. A tool that issues an
+    outbound request with no deadline hangs the agent's executor loop until the
+    remote gives up, so the run never completes and never reports why.
+
+rules:
+  - id: LC-007
+    title: LangChain tool network call has no timeout
+    severity: high
+    confidence: 0.85
+    language: python
+    applies_to:
+      - langchain_tool
+    scope: tool
+    match:
+      call_without_kwarg:
+        callees:
+          - requests.get
+          - requests.post
+          - requests.put
+          - requests.delete
+          - requests.patch
+          - requests.head
+          - requests.request
+          - httpx.get
+          - httpx.post
+          - httpx.put
+          - httpx.delete
+          - httpx.patch
+          - httpx.head
+          - httpx.request
+          - requests.Session.get
+          - requests.Session.post
+          - urllib.request.urlopen
+          # Bare `urlopen` covers the from-import form.
+          - urlopen
+          # Aliased aiohttp sessions canonicalize to aiohttp.ClientSession.<m>.
+          - aiohttp.ClientSession.get
+          - aiohttp.ClientSession.post
+        missing: timeout
+    explanation: >
+      This LangChain tool makes a network request with no timeout, so it hangs
+      indefinitely on a slow or unresponsive remote. The tool runs synchronously
+      inside the AgentExecutor loop: a stalled request blocks that iteration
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI, this is a genuinely well built pair. The gap is real (LangChain was the only pack with tool discovery and no network-timeout rule), and the execution is careful: the `LC-007` callee list is byte-for-byte the PYD-006 list, severity and confidence line up with the neighbors (`0.85` matching PYD-006, `0.6` matching OAI-016 and VAI-011), both predicates already exist in the engine so correctly no `schema_version` bump, and the explanation text is properly LangChain-framed rather than copied wholesale, including the LC-102 `max_iterations` distinction, which is exactly right. I verified the fixture mirror in trustabl/trustabl#127 is byte-identical to this file.

One coordination note, not a mark against the PR: `LC-007` is also claimed by the open pair trustabl/trustabl-rules#115 and trustabl/trustabl#191. Whichever pair lands first keeps the ID; the other will need to renumber and rebase. `LC-015` has no competing claim.

Separately, every shipped rule needs a rationale doc in trustabl-rulebook (threat model, citations, severity defense). That is not on you to produce; we can handle it or pair on it before this ships. I will coordinate landing the two halves together with the ID question resolved.

sairenchristianbuerano:
**Selected for merge.** This pair keeps `LC-007` and `LC-015`.

Closing trustabl/agent-reliability-rules#115 + trustabl/agent-reliability-analyzer#191, which claimed `LC-007` for the same rule.

Deciding factor: @jhumel-code asked #115 to widen its callee list to match the sibling network packs (urllib and aiohttp in particular). This PR's list is already **byte-identical to `pydantic_ai/network.yaml` on main** — requests, httpx, `urllib.request.urlopen`, bare `urlopen` and the aiohttp ClientSession pair. The change requested there would have produced this file.

sairenchristianbuerano:
**Correction to my previous comment.**

trustabl/agent-reliability-rules#115 merged first, so `LC-007` is now taken on `main` and this PR can no longer ship it — it would be a duplicate id and the pack validator will reject it.

`LC-015` in this PR has no competing claim and is unaffected.

The substance of the earlier comment still holds: this callee list is byte-identical to `pydantic_ai/network.yaml` on main and covers urllib, bare `urlopen` and the aiohttp pair, which the merged version does not. Worth landing that widening as a follow-up against the shipped `LC-007` rather than losing it.

sairenchristianbuerano (changes requested):
Thank you for the contribution, this is very helpful. Please fix the merge conflict.

PR #53, opened 2026-08-24
feat(crewai): add CREW-007, tool network call has no timeout

bradAGI:
CrewAI had no network-timeout rule, unlike Pydantic AI (PYD-006), AutoGen (AG2-012), and the Vercel AI SDK (VAI-011).

The blast radius is larger here than in a single-agent SDK. A crew runs tasks in sequence and threads each task's output forward as context, so a tool stalled inside one task blocks every task behind it and the crew produces nothing. CREW-110's `max_iter` doesn't help — it bounds how many reasoning steps an agent takes, not how long a single tool call may run.

Same `call_without_kwarg` callee list as PYD-006, so the alias-aware `requests.Session` / `aiohttp.ClientSession` handling matches the shipped rules.

```
$ trustabl rules validate .
```

Fire (`requests.get(url)`): `CREW-005, CREW-007, CREW-201`
Silent (`requests.get(url, timeout=10)`): `CREW-005, CREW-201`

### DIFF
--- a/crewai/network.yaml
+++ b/crewai/network.yaml
@@ -1,0 +1,57 @@
+policy:
+  id: crewai_network
+  name: CrewAI tool network hygiene
+  category: crewai
+  description: >
+    Network-call hygiene inside CrewAI tool functions. A tool that issues an
+    outbound request with no deadline stalls the agent that called it, and in a
+    crew that agent's task blocks every task sequenced behind it.
+
+rules:
+  - id: CREW-007
+    title: CrewAI tool network call has no timeout
+    severity: high
+    confidence: 0.85
+    language: python
+    applies_to:
+      - crewai_tool
+    scope: tool
+    match:
+      call_without_kwarg:
+        callees:
+          - requests.get
+          - requests.post
+          - requests.put
+          - requests.delete
+          - requests.patch
+          - requests.head
+          - requests.request
+          - httpx.get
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI, this is a clean gap fill and exactly the parity the CrewAI pack was missing. I verified CREW-007 against the engine at main: the schema is complete, `crewai_tool` is a valid `applies_to` for `scope: tool`, `call_without_kwarg` with `callees` plus `missing` is a shipped schema 14 predicate, so no `schema_version` bump is needed, and no other open PR claims the CREW-007 ID. The callee list is byte-identical to PYD-006, which keeps the alias-aware `requests.Session` and `aiohttp.ClientSession` handling consistent across packs, and severity high with confidence 0.85 matches that neighbor exactly. The explanation is genuinely CrewAI-framed rather than copied text, and the CREW-110 cross-reference is accurate (that rule is the `max_iter` cap, which indeed bounds reasoning steps, not tool wall clock). The engine half, trustabl/trustabl#128, mirrors this file byte for byte on a branch of the same name, so the sync check will compare the pair correctly; I will land the two together. One item on our side: every shipped rule needs a rationale doc in trustabl-rulebook, which external PRs understandably do not include. We can handle that or pair on it, it is not on you. Approving.

PR #54, opened 2026-08-24
feat(pydantic_ai): add PYD-008, tool raises without a ModelRetry path

bradAGI:
Pydantic AI was the last Python pack with tool discovery and no error-contract rule — Claude SDK has CSDK-005, OpenAI OAI-008, ADK has ADK-005, MCP has MCP-006.

The framing is SDK-specific rather than a copy of CSDK-005, because Pydantic AI splits failure into two paths that behave very differently. `ModelRetry` is captured by the agent, appended to the message history as feedback, and re-prompts the model within its `retries` budget. Every other exception propagates out of `agent.run()` and ends the run — so a transient failure the model could have corrected kills the whole call, and the caller gets a traceback that may carry paths or connection strings instead of a result.

The `not: has_body_text: [ModelRetry]` clause is a **false-positive suppressor, not the detection signal**: a tool whose only raise is `raise ModelRetry(...)` is already using the SDK's recovery channel correctly and should stay silent. `ModelRetry` is a bare imported name with no structural signature, which is the narrow case the schema reserves `has_body_text` for. Noted inline in the rule.

```
$ trustabl rules validate .
```

| fixture | findings |
|---|---|
| `raise ValueError(...)`, no try/except | `PYD-008, PYD-101, PYD-201` |
| try/except returning a structured error | `PYD-101, PYD-201` |
| `raise ModelRetry(...)` | `PYD-101, PYD-201` |

### DIFF
--- a/pydantic_ai/error_handling.yaml
+++ b/pydantic_ai/error_handling.yaml
@@ -1,0 +1,51 @@
+policy:
+  id: pydantic_ai_error_handling
+  name: Pydantic AI error contract hygiene
+  category: pydantic_ai
+  description: >
+    Rules covering how a Pydantic AI tool surfaces failure. Pydantic AI draws a
+    hard line between the two: ModelRetry is fed back to the model as a
+    correction it can act on, while any other exception aborts the whole run.
+    A tool that raises the wrong kind turns a recoverable failure into a dead
+    run.
+
+rules:
+  - id: PYD-008
+    title: Pydantic AI tool raises without a ModelRetry or structured error path
+    severity: low
+    confidence: 0.6
+    language: python
+    applies_to:
+      - pydantic_ai_tool
+    scope: tool
+    # has_raise + no try/except is the same shape as CSDK-005 and MCP-006. The
+    # `not: has_body_text` clause is a suppressor, not the detection signal: a
+    # tool whose only raise IS `raise ModelRetry(...)` is already using the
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI, this is a well-made rule. You are right that `pydantic_ai` was the last Python pack with tool discovery and no error-contract rule, and the framing is genuinely SDK-specific rather than a reword of CSDK-005: the `ModelRetry` versus propagating-exception split is exactly the distinction that matters here, and both the explanation and fix text reflect it. Severity and confidence (low, 0.6) match CSDK-005, the closest analogue. The `has_body_text` use is the part I looked at hardest, since the schema marks it a discouraged last resort, but you used it correctly: as a false-positive suppressor rather than a detection signal, so its brittleness only errs toward staying silent (a tool that mentions `ModelRetry` in a comment slips through, which is the acceptable direction), and the inline comment documents the reasoning. `PYD-008` is the next free ID and no other open PR claims it. The engine half in trustabl/trustabl#130 carries a byte-identical mirror plus fire and silent coverage, so the pair merges cleanly together. One note on process rather than on your work: every shipped rule needs a rationale doc in the trustabl-rulebook repo before it goes out, and we can handle that side or pair on it with you. Approving, to land together with the engine half.

sairenchristianbuerano (approved):

PR #55, opened 2026-08-24
feat(langchain): add LC-008, tool raises without a structured error contract

bradAGI:
LangChain had no error-contract rule. Claude SDK (CSDK-005), OpenAI (OAI-008), ADK (ADK-005), and MCP (MCP-006) all ship one.

The framing is LangChain-specific. With the default `handle_tool_error=False`, a tool exception propagates out of the `AgentExecutor` and aborts the run — the model is never told the step failed, and the caller gets a traceback instead of an answer. Opting in with `handle_tool_error=True` is only marginally better: the model receives the raw `str(exception)`, with no failure mode, no retryability signal, and whatever internal detail the message happened to carry. The fix text points at `handle_tool_error` as a callable rather than `True`, which is the actual remediation.

**Python only, deliberately.** I drafted a LangChain.js counterpart and dropped it after testing: `PredHasRaise` matches the Python grammar node `raise_statement`, so `has_raise` is structurally always false for a TypeScript tool (whose throws parse as `throw_statement`). The rule validated cleanly but never fired on a fixture that throws. That's an engine-side gap worth closing — `PredHasShellCall` already branches on `models.IsTSOrJS` and reads a discovery-computed fact, so `PredHasRaise` could do the same — but it needs to land in the engine before a TS error-contract rule in this pack (or any other) is worth shipping. Happy to open that PR if useful.

```
$ trustabl rules validate .
```

Fire (`raise ValueError`, no try/except): `LC-008, LC-201`
Silent (try/except returning a structured error): `LC-201`

### DIFF
--- a/langchain/error_handling.yaml
+++ b/langchain/error_handling.yaml
@@ -1,0 +1,39 @@
+policy:
+  id: langchain_error_handling
+  name: LangChain error contract hygiene
+  category: langchain
+  description: >
+    Rules covering how a LangChain tool surfaces failure. By default an
+    exception raised inside a tool escapes the AgentExecutor and ends the run,
+    so the model never sees the failure and never gets a chance to recover from
+    it.
+
+rules:
+  - id: LC-008
+    title: LangChain tool raises without a structured error contract
+    severity: low
+    confidence: 0.6
+    language: python
+    applies_to:
+      - langchain_tool
+    scope: tool
+    match:
+      all:
+        - has_raise: true
+        - has_try_except: false
+    explanation: >
+      This tool raises an exception it never catches. LangChain does not feed
+      tool exceptions back to the model unless you opt in: with the default
+      handle_tool_error=False the exception propagates out of the AgentExecutor
+      and aborts the run, so the model is never told the step failed and the
+      caller gets a traceback instead of an answer. Opting in with
+      handle_tool_error=True is barely better — the model receives the raw
+      str(exception), an opaque sentence with no failure mode, no indication
+      whether a retry could succeed, and whatever internal detail the message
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI, this is a well-built rule. Schema-wise it is clean: all required fields are present, `language: python` is explicit, `langchain_tool` is a valid tool-scope token, and both `has_raise` and `has_try_except` already exist in the engine, so you are right that no `schema_version` bump is needed. Severity low and confidence 0.6 line up exactly with the sibling error-contract rules `CSDK-005` and `MCP-006`, and the explanation is genuinely LangChain-framed rather than copied prose; the point that `handle_tool_error=True` only hands the model the raw `str(exception)` and that the real remediation is a callable is accurate, and that is exactly the framing we want. The engine half in trustabl/trustabl#132 mirrors this file byte for byte and carries the fire and silent cases, so the sync contract is satisfied as long as the two merge together, and the shared branch name for the sync job is appreciated. Two coordination notes. First, `LC-008` is also claimed by trustabl/trustabl#122, trustabl/trustabl#123, trustabl/trustabl-rules#65, and trustabl/trustabl-rules#77, so whichever pair lands first, the others will need to renumber and rebase. Second, before this ships it needs a rationale doc in trustabl-rulebook; we can handle that or pair on it, so it is not something you need to block on.

sairenchristianbuerano:
**Selected for merge.** This pair keeps `LC-008`.

Closing trustabl/agent-reliability-rules#65 (same rule, different author) and trustabl/agent-reliability-rules#77 (used `LC-008` for the idempotency rule instead).

sairenchristianbuerano (approved):

PR #56, opened 2026-08-24
feat(crewai): add CREW-008, tool raises without a structured error contract

bradAGI:
CrewAI had no error-contract rule. Claude SDK (CSDK-005), OpenAI (OAI-008), ADK (ADK-005), and MCP (MCP-006) all ship one.

Framed for how CrewAI actually behaves. An uncaught tool exception is stringified into the agent's observation, so the agent sees a bare message — no failure mode, no signal about whether a retry could succeed — and typically burns its remaining `max_iter` budget re-calling the same tool the same way. The damage compounds in a crew: each task's output is threaded forward as context, so an unusable error string becomes the input every downstream task reasons over.

```
$ trustabl rules validate .
```

Fire (`raise ValueError`, no try/except): `CREW-008, CREW-201`
Silent (try/except returning an actionable error string): `CREW-201`

### DIFF
--- a/crewai/error_handling.yaml
+++ b/crewai/error_handling.yaml
@@ -1,0 +1,39 @@
+policy:
+  id: crewai_error_handling
+  name: CrewAI error contract hygiene
+  category: crewai
+  description: >
+    Rules covering how a CrewAI tool surfaces failure. A tool that raises hands
+    the agent a stringified exception in place of a result, and in a sequential
+    crew that unusable output is threaded forward as context to every task
+    behind it.
+
+rules:
+  - id: CREW-008
+    title: CrewAI tool raises without a structured error contract
+    severity: low
+    confidence: 0.6
+    language: python
+    applies_to:
+      - crewai_tool
+    scope: tool
+    match:
+      all:
+        - has_raise: true
+        - has_try_except: false
+    explanation: >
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI, this fills a real gap. CSDK-005, OAI-008, ADK-005, and MCP-006 all exist and CrewAI had nothing on the error contract, so CREW-008 is a natural addition. I checked the schema fields, the `crewai_tool` applies_to token, and that `has_raise` and `has_try_except` are both shipped predicates, so no `schema_version` bump is correct. Severity low and confidence 0.6 line up exactly with CSDK-005 and OAI-008. The explanation and fix are properly CrewAI-framed (the stringified observation, the `max_iter` burn, the crew threading the string forward as downstream context) rather than lifted from another pack, which is exactly what we ask for.

The engine half in trustabl/trustabl#133 mirrors this file byte for byte and adds fire and silent cases, so the pair satisfies the sync contract. We will merge this half first so the engine PR's rules-sync check goes green against main. Two small notes, neither blocking: CREW-008 skips CREW-007, which I assume echoes OAI-008 (the CREW range already has gaps, so fine), and a shipped rule needs a rationale doc in trustabl-rulebook before release; we can handle or pair on that part, it is not on you. Approving.

PR #57, opened 2026-08-24
feat(autogen): add AG2-013, tool raises without a structured error contract

bradAGI:
AutoGen had no error-contract rule. Claude SDK (CSDK-005), OpenAI (OAI-008), ADK (ADK-005), and MCP (MCP-006) all ship one.

Framed for AutoGen's two-agent reply loop rather than copied from CSDK-005. Tools execute on the executor agent's side and the result is sent back as that agent's reply, so an uncaught exception either aborts `initiate_chat` outright or lands in the message history as a stringified traceback. The second case is the quiet one: both agents keep conversing with a traceback as the last observation, the assistant has no failure mode to branch on, and it re-issues the same call until AG2-004's `max_round` or AG2-006's auto-reply cap ends the chat. The traceback also stays in the transcript, carrying file paths into every subsequent turn's context.

Used AG2-013 rather than the free AG2-003 — that slot looks retired and IDs are external identifiers, so reuse seemed like the wrong call. Happy to renumber if you'd prefer the gap filled.

```
$ trustabl rules validate .
```

Fire (`raise ValueError`, no try/except, registered via `register_for_llm`/`register_for_execution`): `AG2-013, AG2-201`
Silent (try/except returning an actionable error string): `AG2-201`

### DIFF
--- a/autogen/error_handling.yaml
+++ b/autogen/error_handling.yaml
@@ -1,0 +1,41 @@
+policy:
+  id: autogen_error_handling
+  name: AutoGen error contract hygiene
+  category: autogen
+  description: >
+    Rules covering how a registered AutoGen tool surfaces failure. An
+    unhandled exception in a tool breaks the two-agent reply loop the whole
+    conversation is built on, so the failure surfaces as a stalled or aborted
+    chat rather than something either agent can reason about.
+
+rules:
+  - id: AG2-013
+    title: AutoGen tool raises without a structured error contract
+    severity: low
+    confidence: 0.6
+    language: python
+    applies_to:
+      - autogen_tool
+    scope: tool
+    match:
+      all:
+        - has_raise: true
+        - has_try_except: false
+    explanation: >
+      This registered tool raises an exception it never catches. AutoGen
+      executes tools on the executor agent's side of the conversation and sends
+      the result back as that agent's reply, so an uncaught exception either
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI, this fills a real gap. The error-contract family (CSDK-005, OAI-008, ADK-005, MCP-006) had no AutoGen member, and you matched the family exactly where it counts, `severity: low`, `confidence: 0.6`, and the same `has_raise` plus `has_try_except: false` match, while the explanation is genuinely reframed for the two-agent reply loop rather than copied prose. Prescribing an actionable message instead of the `{"error": ..., "retryable": ...}` dict is the right call for AutoGen, since the assistant reasons over the executor's reply, not a typed result. Skipping the retired AG2-003 slot was also correct, shipped IDs never get reused, so please do not renumber. I verified `autogen_tool` is a valid tool-scope token, both predicates exist in the engine at schema 14 (so no version bump is needed), and no other open PR claims AG2-013.

The engine half in trustabl/trustabl#134 mirrors this file byte for byte with a solid fire and silent case, so the sync contract is satisfied. One note on our side: every shipped rule needs a rationale doc in the trustabl-rulebook repo before it goes out. We can handle that or pair on it, it is not a demand on you. I will merge this first and the engine half right behind it, and CI regenerates the attestation files after merge, so nothing more is needed here.

PR #58, opened 2026-08-24
feat(pydantic_ai): add PYD-009, path parameter used in I/O without validation

bradAGI:
Pydantic AI had no path-safety rule. Claude SDK (CSDK-004), OpenAI (OAI-006), ADK (ADK-004), and MCP (MCP-005) all ship one.

The SDK-specific point — and the reason this isn't just CSDK-004 with the names swapped — is that Pydantic AI's type layer looks like it covers this and doesn't. A parameter annotated `str` or `Path` validates cleanly while still carrying `../../etc/passwd`, because the annotation constrains the value's *type*, not the region of the filesystem it points at. The fix text therefore pushes containment into the parameter's type (an `AfterValidator` or a custom path type) so it holds for every tool that accepts it, rather than being re-implemented at each call site.

Uses the per-param `call_uses_unnormalized_path_param` with the same callee set as CSDK-004, so a tool with two path params and one `.resolve()` still fires on the unresolved one.

```
$ trustabl rules validate .
```

Fire (`open(note_path)` on a raw param): `PYD-009, PYD-101, PYD-201`
Silent (`Path(note_path).resolve()` + `is_relative_to` root check): `PYD-101, PYD-201`

### DIFF
--- a/pydantic_ai/path_safety.yaml
+++ b/pydantic_ai/path_safety.yaml
@@ -1,0 +1,47 @@
+policy:
+  id: pydantic_ai_path_safety
+  name: Pydantic AI filesystem path safety
+  category: pydantic_ai
+  description: >
+    Rules that catch a model-supplied filesystem path flowing into an I/O call
+    inside a Pydantic AI tool without containment. A traversal payload like
+    ../../etc/passwd reaches real filesystem state if the tool does not resolve
+    the path and check it against an allowed root.
+
+rules:
+  - id: PYD-009
+    title: Pydantic AI tool uses a path parameter in I/O without validation
+    severity: high
+    confidence: 0.7
+    language: python
+    applies_to:
+      - pydantic_ai_tool
+    scope: tool
+    # call_uses_unnormalized_path_param is per-param: a tool with two path
+    # params and one .resolve() correctly fires on the unresolved one. A
+    # body-wide check would suppress the finding whenever any normalization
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI, this fills a real gap. Pydantic AI was missing a path safety rule that every comparable pack already ships, and the framing is genuinely SDK specific rather than CSDK-004 with the names swapped. The point that a `str` or `Path` annotation validates a traversal payload cleanly is the right hook, and the fix prescribing an `AfterValidator` or a custom path type is the correct Pydantic-native answer instead of repeating the check at each call site. Schema checks all pass against main: every required field is present, `pydantic_ai_tool` is valid for `scope: tool`, `call_uses_unnormalized_path_param` exists in the engine with the `callees` and `callee_prefixes` shape you use, and severity high with confidence 0.7 matches CSDK-004 exactly, which is the right calibration given it is the same predicate. `PYD-009` is free on main, and you correctly steered around `PYD-008`, which the open pair trustabl/trustabl#130 and trustabl/trustabl-rules#54 already claims. The engine mirror in trustabl/trustabl#136 is byte identical and its fire and silent cases hold up against the predicate, so the two-repo contract is satisfied; the two halves should merge together. One note before this ships: every rule needs a rationale doc in trustabl-rulebook (threat model, citations, severity defense). That is not a demand on you, we can handle it or pair on it. Approving.

PR #60, opened 2026-08-24
feat(crewai): add CREW-009, path parameter used in I/O without validation

bradAGI:
CrewAI had no path-safety rule. Claude SDK (CSDK-004), OpenAI (OAI-006), ADK (ADK-004), and MCP (MCP-005) all ship one.

Two CrewAI-specific points shape the text. The `args_schema` looks like it covers this and doesn't — a Pydantic field typed `str` or `Path` validates cleanly while still carrying `../../etc/passwd`, because the schema constrains the argument's *type*, not the region of the filesystem it points at. And the reach is wider than the agent that owns the tool: a crew shares one process and threads each task's output forward as context, so a file read through this tool lands in the transcript every downstream task reasons over — and CREW-104's delegation means an agent never granted the tool can still reach it by asking a peer.

Uses the per-param `call_uses_unnormalized_path_param` with the same callee set as CSDK-004.

```
$ trustabl rules validate .
```

Fire (`open(note_path)` on a raw param): `CREW-009, CREW-201`
Silent (`Path(note_path).resolve()` + `is_relative_to` root check): `CREW-201`

### DIFF
--- a/crewai/path_safety.yaml
+++ b/crewai/path_safety.yaml
@@ -1,0 +1,49 @@
+policy:
+  id: crewai_path_safety
+  name: CrewAI filesystem path safety
+  category: crewai
+  description: >
+    Rules that catch a model-supplied filesystem path flowing into an I/O call
+    inside a CrewAI tool without containment. A traversal payload like
+    ../../etc/passwd reaches real filesystem state if the tool does not resolve
+    the path and check it against an allowed root.
+
+rules:
+  - id: CREW-009
+    title: CrewAI tool uses a path parameter in I/O without validation
+    severity: high
+    confidence: 0.7
+    language: python
+    applies_to:
+      - crewai_tool
+    scope: tool
+    # call_uses_unnormalized_path_param is per-param: a tool with two path
+    # params and one .resolve() correctly fires on the unresolved one. A
+    # body-wide check would suppress the finding whenever any normalization
+    # appears, masking real exposure on the other param.
+    match:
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI, this is a well built rule. I verified `CREW-009` is unclaimed on main and by any other open PR, `crewai_tool` is a valid tool-scope token, and `call_uses_unnormalized_path_param` with `callees` plus `callee_prefixes` is exactly the shape the engine supports, using the same callee set as CSDK-004. Severity high with confidence 0.7 also lines up with CSDK-004, which is the right anchor for a path-safety rule. The explanation is genuinely CrewAI framed rather than ported: the point that `args_schema` constrains the argument's type and not where it points is accurate, and the shared-process, forwarded-context, delegation reach argument earns the severity. The fix prescribing resolve plus an allowed-root check, encoded as a validator on the `args_schema` field, is the right concrete advice. No new predicates, so leaving `schema_version` alone is correct.

The engine half at trustabl/trustabl#139 mirrors this file byte for byte and carries the fire and silent cases, so the sync contract holds as long as the two land together, and I will merge them as a pair. One note that is not on you: every shipped rule needs a rationale doc in the trustabl-rulebook repo before release, and we can handle that side or pair on it. One tiny nit, take it or leave it: referencing CREW-104 by ID inside the explanation reads a bit inside-baseball in a scan report, but I would not hold the PR for that.

PR #61, opened 2026-08-24
feat(autogen): add AG2-014, path parameter used in I/O without validation

bradAGI:
AutoGen had no path-safety rule. Claude SDK (CSDK-004), OpenAI (OAI-006), ADK (ADK-004), and MCP (MCP-005) all ship one.

The AutoGen-specific point is worth stating in the finding text, because it's a mitigation people reasonably assume they already have: **a registered tool sidesteps the containment AG2-001 is about.** Generated code may run inside a Docker executor, but a function registered with `register_for_execution` runs in the host process with the host's filesystem view no matter how the code executor is configured. So "we run code in Docker" is not an answer to this finding, and the fix has to live inside the registered function — it's the only boundary the tool path crosses. Whatever the tool reads is then posted into the message history both agents keep reasoning over.

Type hints don't close it either: a parameter annotated `str` or `Path` passes AutoGen's schema generation while still carrying a traversal payload.

Uses the per-param `call_uses_unnormalized_path_param` with the same callee set as CSDK-004.

```
$ trustabl rules validate .
```

Fire (`open(note_path)` on a raw param): `AG2-014, AG2-201`
Silent (`Path(note_path).resolve()` + `is_relative_to` root check): `AG2-201`

### DIFF
--- a/autogen/path_safety.yaml
+++ b/autogen/path_safety.yaml
@@ -1,0 +1,50 @@
+policy:
+  id: autogen_path_safety
+  name: AutoGen filesystem path safety
+  category: autogen
+  description: >
+    Rules that catch a model-supplied filesystem path flowing into an I/O call
+    inside a registered AutoGen tool without containment. A traversal payload
+    like ../../etc/passwd reaches real filesystem state if the tool does not
+    resolve the path and check it against an allowed root.
+
+rules:
+  - id: AG2-014
+    title: AutoGen tool uses a path parameter in I/O without validation
+    severity: high
+    confidence: 0.7
+    language: python
+    applies_to:
+      - autogen_tool
+    scope: tool
+    # call_uses_unnormalized_path_param is per-param: a tool with two path
+    # params and one .resolve() correctly fires on the unresolved one. A
+    # body-wide check would suppress the finding whenever any normalization
+    # appears, masking real exposure on the other param.
+    match:
+      call_uses_unnormalized_path_param:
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI, this is a well built rule. The gap is real (there is no `autogen/path_safety.yaml` on main, while `claude_sdk`, `openai_sdk`, `adk`, and `mcp` all ship one), and the AutoGen-specific framing is the strongest part: the point that a function registered with `register_for_execution` runs in the host process with the host's filesystem view regardless of the code executor's Docker config is exactly right, and it is the mitigation users would otherwise assume covers them. I verified `call_uses_unnormalized_path_param` exists on the engine at schema 14 with the `callees` plus `callee_prefixes` shape you used, `autogen_tool` is valid for `scope: tool`, `AG2-014` is not claimed by any other open PR, the `AG2-001` cross-reference is accurate, and severity high with confidence 0.7 matches CSDK-004, which uses the identical callee set. So no schema bump needed, as you said. One wording nit: the explanation says "a crew may run its generated code inside a Docker executor", but crew is CrewAI vocabulary; for AutoGen this should be a team, or just the agents. Since the fixture mirror in trustabl/trustabl#140 must stay byte identical, the same one-word fix needs to land there too. One note for our side rather than yours: `AG2-014` will need a rationale doc in the rulebook repo before it ships; we can handle that or pair on it.

PR #62, opened 2026-08-24
feat(langchain): add LC-017, mutating tool has no idempotency key

bradAGI:
LangChain had no idempotency rule. CrewAI (CREW-006), Pydantic AI (PYD-007), MCP (MCP-007), Claude SDK (CSDK-006), and OpenAI (OAI-009) all ship one.

The LangChain-specific argument is that **repetition is routine here, not exceptional.** An `AgentExecutor` step whose observation reads as inconclusive — a timeout, a truncated response, an error string returned under `handle_tool_error` — leads the model to re-reason about the same goal and re-issue the same action, and it has no way to know the first call already committed. `with_retry` and external retry wrappers compound it: a network failure *after* the side effect landed is indistinguishable from one before.

The fix text adds a point the other packs' rules leave implicit — the key has to be derived from the unit of work, not generated fresh per call, or it de-duplicates nothing.

Same `name_has_prefix` + `param_name_matches` shape as MCP-007, so the detection behavior is consistent across packs.

```
$ trustabl rules validate .
```

Fire (`refund_payment(charge_id, amount_cents)`): `LC-017, LC-201`
Silent (same tool with an `idempotency_key` param): `LC-201`

### DIFF
--- a/langchain/idempotency.yaml
+++ b/langchain/idempotency.yaml
@@ -1,0 +1,53 @@
+policy:
+  id: langchain_idempotency
+  name: LangChain mutating-tool idempotency
+  category: langchain
+  description: >
+    Rules that flag mutating LangChain tools (create/send/refund/...) with no
+    idempotency key. A ReAct agent re-calls a tool whenever an observation reads
+    as inconclusive, so a mutation with no key can commit twice.
+
+rules:
+  - id: LC-017
+    title: Mutating LangChain tool has no idempotency key
+    severity: medium
+    confidence: 0.55
+    language: python
+    applies_to:
+      - langchain_tool
+    scope: tool
+    match:
+      all:
+        - name_has_prefix:
+            - create_
+            - send_
+            - delete_
+            - post_
+            - update_
+            - refund_
+            - charge_
+            - issue_
+        - not:
+            param_name_matches:
+              contains:
+                - idempot
+              exact:
+                - request_id
+                - txn_id
+    explanation: >
+      The tool name signals a side effect (create/send/refund/…) and the tool
+      takes no idempotency key, so a repeated call commits the mutation twice.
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI, this is a well made rule. I verified `LC-017` is unclaimed on main and by any other open PR, `langchain_tool` is a valid tool-scope `applies_to`, and every predicate (`name_has_prefix`, `param_name_matches` with `contains`/`exact`) exists in the current engine schema, so you are right that no `schema_version` bump is needed. The calibration is exactly right too: severity `medium` and confidence `0.55` match MCP-007, whose shape you deliberately mirrored, and the pack-level `policy` block follows the same pattern as `mcp_idempotency`. The explanation is genuinely LangChain-framed rather than copied text (the AgentExecutor inconclusive-observation loop and `with_retry` argument), and the fix text's point about deriving the key from the unit of work instead of generating one per call is a real improvement over the sibling rules.

The engine half in trustabl/trustabl#141 mirrors this file byte for byte (same blob hash) and carries fire plus silent cases, so the sync contract is satisfied and this will not break `rules-sync` as long as the two land together, which I will coordinate. One housekeeping note that is on us, not you: every shipped rule needs a rationale doc in the trustabl-rulebook repo, and we can handle or pair on that before this ships. Happy to merge this alongside the engine PR.

sairenchristianbuerano:
**Selected for merge.** This pair keeps `LC-017`.

The LangChain idempotency rule was submitted three times under three ids — `LC-008` (trustabl/agent-reliability-rules#77), `LC-009` (trustabl/agent-reliability-rules#65) and `LC-017` here. `LC-017` is the only one that collides with nothing, so it wins and the other two close.

sairenchristianbuerano (approved):

PR #63, opened 2026-08-24
feat(autogen): add AG2-015, mutating tool has no idempotency key

bradAGI:
AutoGen had no idempotency rule. CrewAI (CREW-006), Pydantic AI (PYD-007), MCP (MCP-007), Claude SDK (CSDK-006), and OpenAI (OAI-009) all ship one.

AutoGen's conversation loop makes repetition **the default failure mode**, which is a stronger argument than the one the other packs make. The assistant proposes the call, the executor runs it and replies with the result, and if that reply reads as inconclusive — a timeout, an error string, a truncated payload — the assistant proposes the same call again on the next round, with no way to know the first already committed. It keeps doing so until AG2-004's `max_round` or AG2-006's auto-reply cap stops the chat, which **bounds how many duplicates land rather than preventing them**. A group chat is worse: a second agent reading the same inconclusive transcript can re-propose the action independently.

The fix text notes the key must be derived from the unit of work rather than generated per call, or it de-duplicates nothing.

Same `name_has_prefix` + `param_name_matches` shape as MCP-007. AG2-015 continues from AG2-013/014 in the other PRs in this series; AG2-003 is left alone as a retired slot.

```
$ trustabl rules validate .
```

Fire (`refund_payment(charge_id, amount_cents)` registered via `register_for_llm`/`register_for_execution`): `AG2-015, AG2-201`
Silent (same tool with an `idempotency_key` param): `AG2-201`

### DIFF
--- a/autogen/idempotency.yaml
+++ b/autogen/idempotency.yaml
@@ -1,0 +1,54 @@
+policy:
+  id: autogen_idempotency
+  name: AutoGen mutating-tool idempotency
+  category: autogen
+  description: >
+    Rules that flag mutating AutoGen tools (create/send/refund/...) with no
+    idempotency key. A two-agent conversation re-proposes an action whenever the
+    previous observation reads as inconclusive, so a mutation with no key can
+    commit once per round.
+
+rules:
+  - id: AG2-015
+    title: Mutating AutoGen tool has no idempotency key
+    severity: medium
+    confidence: 0.55
+    language: python
+    applies_to:
+      - autogen_tool
+    scope: tool
+    match:
+      all:
+        - name_has_prefix:
+            - create_
+            - send_
+            - delete_
+            - post_
+            - update_
+            - refund_
+            - charge_
+            - issue_
+        - not:
+            param_name_matches:
+              contains:
+                - idempot
+              exact:
+                - request_id
+                - txn_id
+    explanation: >
+      The tool name signals a side effect (create/send/refund/…) and the tool
+      takes no idempotency key, so a repeated call commits the mutation twice.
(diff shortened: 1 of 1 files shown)


jhumel-code:
@bradAGI thanks for this one, it is in good shape. Every required schema field is present, `autogen_tool` is valid at tool scope, and both predicates (`name_has_prefix`, `param_name_matches`) already exist in the engine, so you are right that no `schema_version` bump is needed. Severity and confidence line up exactly with the sibling MCP-007, and the explanation is genuinely AutoGen framed rather than copied text; the point that AG2-004's `max_round` and AG2-006's auto-reply cap bound duplicates rather than prevent them is accurate against the shipped pack. The engine half in trustabl/trustabl#142 mirrors this file byte for byte and carries fire plus silent cases, so the pair satisfies the sync contract.

The one thing to sort before merge is coordination. AG2-015 is also claimed by your other open pair, trustabl/trustabl#138 and trustabl/trustabl-rules#101, which add the same rule with different prose. Both cannot land, so please pick one pair and close the other; for what it is worth I prefer this version's explanation. The gap after AG2-012 is fine given AG2-013/014 sit in your other series PRs. Last note, every shipped rule needs a rationale doc in trustabl-rulebook before it ships; we can handle or pair on that part, it is not on you to include here.

sairenchristianbuerano:
**Closing in favour of trustabl/agent-reliability-rules#101 + trustabl/agent-reliability-analyzer#138**, which is the same `AG2-015` rule you submitted on a second branch — identical title, severity, confidence and match expression. That pair merges without conflicts, this one needs a rebase, so we kept the clean one. Nothing is lost.

PR #64, opened 2026-08-24
feat(pydantic_ai): add PYD-010, PYD-011 tool description quality rules

bradAGI:
Ports the CSDK-017/018 description-quality pair (merged in #40) to Pydantic AI. PYD-001 only checks that a docstring *exists*, so a tool whose docstring reads `"""TODO: describe this tool."""` or `"""Gets data."""` passes today while giving the model no selection signal at all.

The reason this is worth more than a straight port: **the cost is higher in this pack than in the Claude SDK**, and the rule text says so. Pydantic AI doesn't just forward the docstring as the tool description — it also parses the docstring's parameter sections (Google/NumPy/Sphinx) into the argument schema. So a placeholder or one-clause docstring strips the per-argument guidance at the same time as the tool-level guidance, and nothing in the pipeline reports the omission. The fix text asks for the parameter section explicitly rather than just "a longer description".

Both rules go into the existing `pydantic_ai/tool_definition.yaml` alongside PYD-001/002, which is where the file's policy description already scopes them.

```
$ trustabl rules validate .
```

Fire (one tool with `"""TODO: describe this tool."""`, one with `"""Gets data."""`): `PYD-010, PYD-011, PYD-101, PYD-201`
Silent (full description with an `Args:` section): `PYD-101, PYD-201`

Note PYD-011 pairs `description_length_lt: 40` with `has_docstring: true` so it doesn't double-report against PYD-001 on an empty docstring, same as CSDK-018.

### DIFF
--- a/pydantic_ai/tool_definition.yaml
+++ b/pydantic_ai/tool_definition.yaml
@@ -62,3 +62,62 @@ rules:
       list[str], a Pydantic model, etc.). Pydantic AI propagates these
       annotations into the schema the model sees, so the model emits
       correctly-shaped arguments and validation stops rejecting them.
+
+  - id: PYD-010
+    title: Pydantic AI tool description is a placeholder
+    severity: low
+    confidence: 0.85
+    language: python
+    applies_to:
+      - pydantic_ai_tool
+    scope: tool
+    match:
+      has_description_text:
+        - todo
+        - tbd
+        - fixme
+        - placeholder
+        - no description
+        - does stuff
+    explanation: >
+      The docstring passes the PYD-001 has-a-description check but carries a
+      placeholder marker instead of real content. Pydantic AI forwards this text
+      to the model verbatim as the tool's selection signal, so a stub like
+      "TODO: describe this tool" is functionally indistinguishable from no
+      description at all and the model still has to guess from the function
+      name. It also costs more here than in most SDKs: Pydantic AI parses the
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI, this is a model rules contribution. I verified both predicates (`has_description_text` and `description_length_lt`) already exist in the engine schema on main, so you are right that no `schema_version` bump is needed, and `pydantic_ai_tool` is a valid tool-scope `applies_to` token. `PYD-010` and `PYD-011` are not claimed by any other open PR, and the severity and confidence values (low at 0.85 and 0.8) match the CSDK-017/018 precedent exactly, which is the right anchor. The explanation text earns the port rather than copying it: the point that Pydantic AI parses the docstring's parameter sections into the argument schema, so a placeholder strips per-argument guidance at the same time as tool-level guidance, is correct and specific to this SDK, and the fix text asking for a Google, NumPy, or Sphinx parameter section follows through on it. The `has_docstring: true` guard on `PYD-011` keeps the missing-docstring case with `PYD-001`, mirroring the Claude pack. The engine half in trustabl/trustabl#143 mirrors this file byte for byte and carries the fire and silent cases, so I will land the two together. One note for completeness, not an ask of you: every shipped rule needs a rationale doc in the trustabl-rulebook repo, and we can handle that part or pair on it before these ship. Approving.

PR #66, opened 2026-08-24
feat(langchain): add LC-018, LC-019 tool description quality rules

bradAGI:
Ports the CSDK-017/018 description-quality pair (merged in #40) to LangChain. LC-001 only checks that a docstring *exists*, so a tool whose docstring reads `"""TODO: describe this tool."""` or `"""Gets data."""` passes today while giving the model no selection signal.

The LangChain framing is its own on both counts:

- **Mis-selection scales badly here.** LangChain agents are routinely handed a dozen or more tools at once, so the model is picking between similarly named neighbors far more often than in a two-tool setup, and the failure is quiet — wrong answers, not errors anyone traces back to the docstring.
- **It costs the run, not just the answer.** Each mis-selected call consumes an iteration against LC-102's `max_iterations` budget, so a run can exhaust its steps re-trying tools it picked badly and return nothing useful.

LC-018's fix also points at the `@tool` decorator's explicit `description=` for cases where the function name is genuinely ambiguous, which is the LangChain-idiomatic escape hatch.

```
$ trustabl rules validate .
```

Fire (one tool with `"""TODO: describe this tool."""`, one with `"""Gets data."""`): `LC-018, LC-019, LC-201`
Silent (full description naming when to prefer the neighboring tool): `LC-201`

LC-019 pairs `description_length_lt: 40` with `has_docstring: true` so it doesn't double-report against LC-001 on an empty docstring, same as CSDK-018.

### DIFF
--- a/langchain/tool_definition.yaml
+++ b/langchain/tool_definition.yaml
@@ -77,3 +77,63 @@ rules:
       Pass a one-sentence description in the tool() config (or the
       DynamicStructuredTool options) naming what the tool does, the inputs it
       expects, and what it returns. Write it for the model, not a human reader.
+
+  - id: LC-018
+    title: LangChain tool description is a placeholder
+    severity: low
+    confidence: 0.85
+    language: python
+    applies_to:
+      - langchain_tool
+    scope: tool
+    match:
+      has_description_text:
+        - todo
+        - tbd
+        - fixme
+        - placeholder
+        - no description
+        - does stuff
+    explanation: >
+      The docstring passes the LC-001 has-a-description check but carries a
+      placeholder marker instead of real content. LangChain derives the tool
+      description from the docstring and puts it in front of the model as the
+      tool's only selection signal, so a stub like "TODO: describe this tool" is
+      functionally indistinguishable from no description at all — the model
+      still has to guess from the function name. The failure is quiet and gets
+      worse with scale: LangChain agents are routinely handed a dozen or more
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI, this is a clean port. `LC-018` and `LC-019` are both free on `main`, no other open PR claims either ID, and `has_description_text` and `description_length_lt` already ship at schema 14, so you are right that no `schema_version` bump belongs here. Severity and confidence match `CSDK-017` and `CSDK-018` exactly, and pairing `description_length_lt: 40` with `has_docstring: true` correctly leaves the empty-description case to `LC-001`. The rewritten framing (the AgentExecutor iteration budget, the `description=` escape hatch on `@tool`) is genuinely LangChain's rather than restated Claude text, which is the part these ports usually miss. The engine half, trustabl/trustabl#145, mirrors this file byte for byte.

Two non-blocking things I would like your read on. The bare `todo` needle will fire on a legitimate tool whose description is about todo lists, which is common enough in this ecosystem; `todo:` would still catch "TODO: describe this tool" without that class of false positive. And the explanation cites `LC-001` and `LC-102` by ID, which no shipped rule does today, and `LC-102` only fires for an AgentExecutor, so a LangGraph reader may never see the rule being pointed at.

A rulebook rationale doc still needs to land before these ship. That part is on us, and I am happy to pair on it.

PR #67, opened 2026-08-24
feat(crewai): add CREW-010, CREW-011 tool description quality rules

bradAGI:
Ports the CSDK-017/018 description-quality pair (merged in #40) to CrewAI. CREW-001 only checks that a docstring *exists*, so a tool whose docstring reads `"""TODO: describe this tool."""` or `"""Gets data."""` passes today while giving the agent no selection signal.

What makes the CrewAI framing its own is that **the consequence travels**. An agent that picks the wrong tool produces a task output that's threaded forward as context to every task behind it, so one undescribed tool degrades work the agent that owns it never touches. CREW-104 delegation compounds it: peers can reach this tool on the strength of the same placeholder. And each mis-selected call spends one of the agent's CREW-110 `max_iter` steps, so the agent can hit its cap and hand the next task a half-finished result rather than an answer.

```
$ trustabl rules validate .
```

Fire (one tool with `"""TODO: describe this tool."""`, one with `"""Gets data."""`): `CREW-010, CREW-011, CREW-201`
Silent (full description naming when to prefer the neighboring tool): `CREW-201`

CREW-011 pairs `description_length_lt: 40` with `has_docstring: true` so it doesn't double-report against CREW-001 on an empty docstring, same as CSDK-018.

### DIFF
--- a/crewai/tool_definition.yaml
+++ b/crewai/tool_definition.yaml
@@ -52,3 +52,60 @@ rules:
       Annotate every parameter with a concrete type (str, int, list[str], a
       Pydantic model, etc.). CrewAI propagates these annotations into the schema the
       model sees, so the model emits correctly-shaped arguments.
+
+  - id: CREW-010
+    title: CrewAI tool description is a placeholder
+    severity: low
+    confidence: 0.85
+    language: python
+    applies_to:
+      - crewai_tool
+    scope: tool
+    match:
+      has_description_text:
+        - todo
+        - tbd
+        - fixme
+        - placeholder
+        - no description
+        - does stuff
+    explanation: >
+      The docstring passes the CREW-001 has-a-description check but carries a
+      placeholder marker instead of real content. CrewAI puts the tool
+      description in front of the agent as its only basis for choosing between
+      the tools it holds, so a stub like "TODO: describe this tool" is
+      functionally indistinguishable from no description at all. In a crew the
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI. This reads as a real port rather than a copy of `CSDK-017` / `CSDK-018`: the explanation earns its CrewAI framing by naming how a mis-selection travels forward through task context and spends `max_iter` steps, and both `CREW-104` and `CREW-110` are real rules on `main`, so those cross-references hold up.

Everything schema-side checks out against `main`. `has_description_text` and `description_length_lt` both already exist at schema version 14, so you are right that no `schema_version` bump is needed, `crewai_tool` is a valid `applies_to` for `scope: tool`, and `CREW-010` / `CREW-011` are free on `main` and do not collide with your other open CrewAI rule PRs. Pairing `has_docstring: true` with `description_length_lt: 40` is the right call, since an absent docstring is length zero and would otherwise double-report against `CREW-001`.

One non-blocking observation: the `placeholder` needle will also match a genuine description such as "Replace placeholder tokens in a template". That behavior is inherited from `CSDK-017`, and I would rather keep the two packs consistent than special-case one of them.

I will take this together with trustabl/trustabl#146, which carries the byte-identical fixture mirror and the test cases, since either half alone fails `rules-sync`. The only thing still outstanding is a rulebook rationale doc for the two new rules before they ship, and we can handle or pair on that from our side.

PR #68, opened 2026-08-24
feat(autogen): add AG2-016, AG2-017 tool description quality rules

bradAGI:
Ports the CSDK-017/018 description-quality pair (merged in #40) to AutoGen. AG2-007 only checks that a docstring *exists*, so a tool whose docstring reads `"""TODO: describe this tool."""` or `"""Gets data."""` passes today while giving the assistant no selection signal.

**How AutoGen fails on a bad guess is what makes this expensive**, and it's the part worth putting in the finding text. A wrongly chosen tool returns something that doesn't answer the assistant's intent; the assistant reads that as an inconclusive step and proposes another call; the pair trades rounds until AG2-004's `max_round` or AG2-006's auto-reply cap ends the chat. The conversation budget goes on a selection problem that no error message points at — nothing in the transcript says "the description was a stub".

Both fixes name `register_for_llm`'s `description=` argument, which overrides the docstring and is the clearer place to put it in this SDK.

```
$ trustabl rules validate .
```

Fire (one tool with `"""TODO: describe this tool."""`, one with `"""Gets data."""`): `AG2-016, AG2-017, AG2-201`
Silent (full description naming when to prefer the neighboring tool): `AG2-201`

AG2-017 pairs `description_length_lt: 40` with `has_docstring: true` so it doesn't double-report against AG2-007 on an empty docstring, same as CSDK-018.

### DIFF
--- a/autogen/tool_definition.yaml
+++ b/autogen/tool_definition.yaml
@@ -54,3 +54,64 @@ rules:
       Annotate every parameter with a concrete type (str, int, list[str], a
       typing.Annotated description, etc.). AutoGen propagates these annotations into
       the schema the model sees, so the model emits correctly-shaped arguments.
+
+  - id: AG2-016
+    title: AutoGen tool description is a placeholder
+    severity: low
+    confidence: 0.85
+    language: python
+    applies_to:
+      - autogen_tool
+    scope: tool
+    match:
+      has_description_text:
+        - todo
+        - tbd
+        - fixme
+        - placeholder
+        - no description
+        - does stuff
+    explanation: >
+      The docstring passes the AG2-007 has-a-description check but carries a
+      placeholder marker instead of real content. AutoGen advertises the tool to
+      the assistant through this text, so a stub like "TODO: describe this tool"
+      is functionally indistinguishable from no description at all and the
+      assistant has to guess from the function name. The way AutoGen fails on a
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI, this is a model rules contribution. Both rules are schema-clean: `autogen_tool` is the right `applies_to` for tool scope, and `has_description_text` and `description_length_lt` both exist in the engine at schema version 14, so you are correct that no `schema_version` bump is needed. Severity and confidence match the CSDK-017/018 precedent exactly (low, 0.85 and 0.8), and I appreciate that the explanation text was actually reframed for AutoGen rather than copied over: the `max_round` and auto-reply-cap failure mode and the `register_for_llm` `description=` fix are the right framing for this SDK. The `has_docstring: true` guard on AG2-017 keeps the missing-docstring case with AG2-007, which is the same discipline CSDK-018 uses. Neither ID collides with anything on main or with any other open PR. The engine half at trustabl/trustabl#147 mirrors this file byte for byte and carries the fire and silent cases, so the pair satisfies the sync contract, and I will land the two together. One housekeeping note on our side: every shipped rule needs a rationale doc in the trustabl-rulebook repo (threat model, citations, severity defense), which external PRs understandably do not include. We can handle that part or pair on it, it is not a blocker for you. Approving.

PR #69, opened 2026-08-24
feat(vercel_ai): add VAI-013, VAI-014 tool description quality rules

bradAGI:
Ports the CSDK-017/018 description-quality pair (merged in #40) to the Vercel AI SDK. VAI-004 only checks that a `description` *exists*, so a tool whose description reads `"TODO: describe this tool."` or `"Gets data."` passes today — while sitting in exactly the state VAI-004 exists to prevent.

**The gap matters more here than in the Python packs, and VAI-004's own text already says why:** this SDK has no docstring fallback, so the `description` string is the entire prompt-side account of the tool. A placeholder tells the model nothing the function name didn't, and nothing in the SDK reports that the description was a stub — the only symptom is a tool the model uses badly or not at all. Each mis-selected call also consumes a step against the VAI-007 tool-loop bound, so a thin description spends the loop budget instead of the run producing an answer.

Both predicates are confirmed working against TypeScript tools, not assumed — `has_description_text` reads `ToolDef.Description`, which for TS is the explicit `description` field.

```
$ trustabl rules validate .
```

Fire (one tool with `description: "TODO: describe this tool."`, one with `"Gets data."`): `VAI-012, VAI-013, VAI-014`
Silent (full description naming when to prefer the neighboring tool): `VAI-012`

(VAI-012 is the pre-existing repo-hygiene rule firing on the fixture having no AGENTS.md.)

VAI-014 pairs `description_length_lt: 40` with `has_docstring: true` so it doesn't double-report against VAI-004 on an empty description, same as CSDK-018.

### DIFF
--- a/vercel_ai/tool_definition.yaml
+++ b/vercel_ai/tool_definition.yaml
@@ -58,3 +58,60 @@ rules:
       field per argument (z.object({ city: z.string(), units: z.enum([...]) })).
       Reserve dynamicTool for the rare case where the input genuinely cannot be
       typed, and even then validate the shape inside execute() before using it.
+
+  - id: VAI-013
+    title: Vercel AI tool description is a placeholder
+    severity: low
+    confidence: 0.85
+    language: typescript
+    applies_to:
+      - vercel_ai_tool
+    scope: tool
+    match:
+      has_description_text:
+        - todo
+        - tbd
+        - fixme
+        - placeholder
+        - no description
+        - does stuff
+    explanation: >
+      The tool sets a description, so it passes VAI-004, but the string is a
+      placeholder rather than real content. That leaves the tool in exactly the
+      state VAI-004 exists to prevent: the Vercel AI SDK has no docstring
+      fallback, so this string is the entire account of the tool in the prompt,
+      and "TODO: describe this tool" tells the model nothing the function name
+      did not. The model skips the tool or calls it with the wrong arguments,
+      and nothing in the SDK reports that the description was a stub — the only
+      symptom is a tool the model uses badly or not at all.
+    fix: >
+      Replace the placeholder with a real description covering what the tool
+      does, what it returns, and when the model should call it rather than a
+      neighboring tool. The SDK passes the string to the model verbatim, so
+      write it for the model rather than a human maintainer.
+
+  - id: VAI-014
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI, this is a well built pair. I checked the YAML here against the fixture mirror in trustabl/trustabl#148 and the two are byte identical, so the sync contract holds when they land together. Schema-wise everything checks out: both predicates (`has_description_text`, `description_length_lt`) already exist in the engine, so skipping the `schema_version` bump is correct, `vercel_ai_tool` is the right `applies_to` for tool scope, and severity and confidence match the CSDK-017/018 siblings exactly. The double-report guard is also right: VAI-004 in this pack matches on `has_docstring: false`, so the `has_docstring: true` leg of VAI-014 cleanly hands the missing-description case back to VAI-004. And the explanation text is genuinely reframed for this SDK (no docstring fallback, the VAI-007 loop-budget point) rather than pasted from the Python rule, with accurate cross-references.

One coordination note: VAI-013 is also claimed by trustabl/trustabl#121 and trustabl/trustabl-rules#102. Whichever pair merges first, the others will need to renumber and rebase; that is not a mark against this PR, which I have no content changes for. Before the rules ship they will need a rationale doc in the rulebook repo; we can handle that on our side or pair on it.

sairenchristianbuerano:
Rebased onto current `main`. This had gone stale behind the schema-16 work and the Vercel/ADK rules that merged this week.

The conflict was in the shared test fixture only — your rule content, severity, confidence and match expressions are unchanged. Fire/silent cases verified passing locally, and the pack is byte-identical to the engine fixture so `rules-sync` stays green.

Nothing needed from you.

sairenchristianbuerano:
For the record, since this PR was involved in an id clash: `VAI-013` had also been claimed by the privileged-tool approval rule in trustabl/agent-reliability-rules#102 / trustabl/agent-reliability-analyzer#121.

**This PR keeps `VAI-013` and `VAI-014`.** We renumbered the other rule to `VAI-018` instead, because this one claims the two ids as a contiguous pair and was opened first in the rules repo. Nothing changed here as a result — no action needed.

Verified: the validator reports 220 rules valid with both pairs applied, no duplicate ids.

sairenchristianbuerano (approved):

PR #71, opened 2026-08-24
feat(pydantic_ai): add PYD-012, tool prints to stdout for diagnostics

bradAGI:
Ports OAI-010 to Pydantic AI — the observability dimension isn't covered in any of the five newer packs.

**The loss is larger in this SDK than in the OpenAI pack, which is what the rule text leads with.** Pydantic AI emits OpenTelemetry spans for each run and tool call, so everything around this `print` is already correlated to a trace — the bare print is the one diagnostic that lands outside it, unattached to the run that produced it. That's a sharper argument than "it disappears from logs," and it points at a better fix: attach the detail to the current span via Logfire or the OTel API, rather than only swapping in a module logger.

The stdio-transport corruption note from OAI-010 carries over, since a Pydantic AI tool served over an MCP stdio transport has the same JSON-RPC interleaving problem.

```
$ trustabl rules validate .
```

Fire (`print(f"looking up {order_id}")` in the tool body): `PYD-012, PYD-101, PYD-201`
Silent (`logger.info("looking up %s", order_id)`): `PYD-101, PYD-201`

### DIFF
--- a/pydantic_ai/observability.yaml
+++ b/pydantic_ai/observability.yaml
@@ -1,0 +1,37 @@
+policy:
+  id: pydantic_ai_observability
+  name: Pydantic AI tool observability hygiene
+  category: pydantic_ai
+  description: >
+    Rules covering how a Pydantic AI tool emits diagnostics. A tool body that
+    prints to stdout writes outside the SDK's instrumentation, so the record
+    reaches neither the model nor the trace the rest of the run is captured in.
+
+rules:
+  - id: PYD-012
+    title: Pydantic AI tool prints to stdout for diagnostics
+    severity: low
+    confidence: 0.65
+    language: python
+    applies_to:
+      - pydantic_ai_tool
+    scope: tool
+    match:
+      has_print_call: true
+    explanation: >
+      The tool body calls print(), which writes to the process's stdout. The
+      model never sees it — only the return value flows back into the agent run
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI, this is a well-built rule and exactly how I want ports across SDK packs done. The framing is genuinely Pydantic AI's own (the OpenTelemetry span argument, and the Logfire/OTel span attachment in the `fix`) rather than a copy of OAI-010's text with the SDK name swapped. Schema-wise it is clean: all required fields present, `pydantic_ai_tool` is valid for `scope: tool`, `has_print_call` exists in the engine at schema 14 so leaving `manifest.yaml` untouched is correct, and `severity: low` with `confidence: 0.65` matches OAI-010, the right calibration anchor. `PYD-012` collides with no other open PR, and I noticed you deliberately left 008 through 011 to the PRs already claiming them, which saves everyone a renumber. The engine half, trustabl/trustabl#150, mirrors this YAML byte for byte and adds the required fire and silent cases, so the pair satisfies the sync contract; I will merge the two together so `rules-sync` never sees a divergent window. One note before this ships: every rule needs a rationale doc in the trustabl-rulebook repo (threat model, citations, severity defense). That is not on you, we can handle it or pair on it. Approving.

PR #72, opened 2026-08-24
feat(langchain): add LC-020, tool prints to stdout for diagnostics

bradAGI:
Ports OAI-010 to LangChain — the observability dimension isn't covered in any of the five newer packs.

**The loss is larger here than in the OpenAI pack, which is what the rule text leads with.** LangChain already threads every tool start, end, and error through its callback system into LangSmith or whichever tracer is configured. So a bare `print` is the one diagnostic that lands *outside* that — detached from the run and the step that produced it, and simply absent the moment anyone debugs from a trace rather than a terminal. The fix therefore names the callback system as well as a module logger.

The stdio-transport corruption note from OAI-010 carries over, since a LangChain tool re-served over an MCP stdio transport has the same JSON-RPC interleaving problem.

```
$ trustabl rules validate .
```

Fire (`print(f"looking up {order_id}")` in the tool body): `LC-020, LC-201`
Silent (`logger.info("looking up %s", order_id)`): `LC-201`

### DIFF
--- a/langchain/observability.yaml
+++ b/langchain/observability.yaml
@@ -1,0 +1,38 @@
+policy:
+  id: langchain_observability
+  name: LangChain tool observability hygiene
+  category: langchain
+  description: >
+    Rules covering how a LangChain tool emits diagnostics. A tool body that
+    prints to stdout writes outside the callback system the rest of the run is
+    recorded through, so the record reaches neither the model nor the trace.
+
+rules:
+  - id: LC-020
+    title: LangChain tool prints to stdout for diagnostics
+    severity: low
+    confidence: 0.65
+    language: python
+    applies_to:
+      - langchain_tool
+    scope: tool
+    match:
+      has_print_call: true
+    explanation: >
+      The tool body calls print(), which writes to the process's stdout. The
+      model never sees it — only the return value flows back into the executor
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI, this is a clean rule and a well-argued one. Checking it against `main`: `langchain_tool` is a valid `applies_to` at `tool` scope, `has_print_call` already exists in the engine and matches the bare `print` callee only, so your `pprint` claim holds, every required field is set, and since no new predicate is involved you are right that `schema_version` stays at 14. `LC-020` is free on `main` and no other open PR claims it, which also keeps you clear of the contested `LC-007` through `LC-009` band. `severity: low` with `confidence: 0.65` matches `OAI-010` exactly, which is the right anchor for a rule whose false positive mode is a deliberate debug print.

The explanation is genuinely reframed around the callback system and the tracer rather than a straight port of the OpenAI text, and it names a concrete consequence, which is what I want from this pack.

This pairs with trustabl/trustabl#151, which mirrors the file byte for byte and adds the coverage cases, so the two repo contract is satisfied. The remaining piece is a rulebook rationale doc for `LC-020` before the rule ships. That is not on you, we can write it or pair on it.

PR #73, opened 2026-08-24
feat(crewai): add CREW-012, tool prints to stdout for diagnostics

bradAGI:
Ports OAI-010 to CrewAI — the observability dimension isn't covered in any of the five newer packs.

**CrewAI makes this worse than a lost log line, which is what the rule text leads with.** A crew running with `verbose=True` is already writing its own narration to stdout, so a tool `print` gets interleaved into run commentary from several agents at once, with nothing marking which agent, task, or tool call emitted it. That's the trap: it looks like working diagnostics in a terminal and becomes unattributable the moment the crew runs anywhere else. The fix names what a module logger buys you here specifically — the record carries its own module and level and lands in the application's log sink rather than the crew's narration stream.

```
$ trustabl rules validate .
```

Fire (`print(f"looking up {order_id}")` in the tool body): `CREW-012, CREW-201`
Silent (`logger.info("looking up %s", order_id)`): `CREW-201`

### DIFF
--- a/crewai/observability.yaml
+++ b/crewai/observability.yaml
@@ -1,0 +1,37 @@
+policy:
+  id: crewai_observability
+  name: CrewAI tool observability hygiene
+  category: crewai
+  description: >
+    Rules covering how a CrewAI tool emits diagnostics. A tool body that prints
+    to stdout writes into the same channel the crew's own verbose narration uses,
+    so the record is neither visible to the agent nor separable from the run
+    commentary around it.
+
+rules:
+  - id: CREW-012
+    title: CrewAI tool prints to stdout for diagnostics
+    severity: low
+    confidence: 0.65
+    language: python
+    applies_to:
+      - crewai_tool
+    scope: tool
+    match:
+      has_print_call: true
+    explanation: >
+      The tool body calls print(), which writes to the process's stdout. The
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI, this is a clean port. `has_print_call` and `crewai_tool` both exist on `main`, every required field is set, and `severity: low` with `confidence: 0.65` lines up exactly with OAI-010, which is the right neighbor to match against. No other CrewAI rule keys on the same predicate, so there is no double fire, and CREW-012 is unclaimed both on `main` and across the other open PRs, so no renumbering is coming for you. You are also right that no `schema_version` bump belongs here, since the rule adds no predicate.

I particularly like that the text is CrewAI's own rather than a copy of the OpenAI framing. The `verbose=True` interleaving point is what justifies a separate rule instead of widening `applies_to` on OAI-010.

The engine half in trustabl/trustabl#152 mirrors this file byte for byte and carries the fire and silent cases, so the pair lands together. One item is on us rather than on you: a shipped rule needs a rationale doc in `trustabl-rulebook` under `docs/Policy/crewai/`, carrying the threat model and the OWASP references. We can write that or pair with you on it, and it is not a blocker here. Nothing to do about `badges/rules.json` or the attestation files either, CI regenerates those after merge.

PR #74, opened 2026-08-24
feat(autogen): add AG2-018, tool prints to stdout for diagnostics

bradAGI:
Ports OAI-010 to AutoGen — the observability dimension isn't covered in any of the five newer packs.

**AutoGen makes this especially deceptive, which is what the rule text leads with.** The conversation itself is narrated to stdout, so a tool `print` lands in the middle of the transcript and *reads as though it were part of the exchange* — when in fact it's invisible to the agents reasoning over it and carries no marker for which round or which tool call produced it. Anyone reconstructing the run from a log sink rather than a terminal loses it entirely.

The fix points at the return value as "the only channel the agents actually read," which is the distinction that matters in a two-agent reply loop.

```
$ trustabl rules validate .
```

Fire (`print(f"looking up {order_id}")` in a tool registered via `register_for_llm`/`register_for_execution`): `AG2-018, AG2-201`
Silent (`logger.info("looking up %s", order_id)`): `AG2-201`

### DIFF
--- a/autogen/observability.yaml
+++ b/autogen/observability.yaml
@@ -1,0 +1,39 @@
+policy:
+  id: autogen_observability
+  name: AutoGen tool observability hygiene
+  category: autogen
+  description: >
+    Rules covering how a registered AutoGen tool emits diagnostics. A tool body
+    that prints to stdout writes into the same stream the conversation itself is
+    narrated on, so the record is neither visible to the agents nor separable
+    from the transcript around it.
+
+rules:
+  - id: AG2-018
+    title: AutoGen tool prints to stdout for diagnostics
+    severity: low
+    confidence: 0.65
+    language: python
+    applies_to:
+      - autogen_tool
+    scope: tool
+    match:
+      has_print_call: true
+    explanation: >
+      The registered tool's body calls print(), which writes to the process's
+      stdout. Neither agent sees it — only the return value is posted back into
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI, this is a model contribution. I checked the rule against the engine at main: `has_print_call` exists in `schema.go`, `predicates.go`, and the evaluator, `autogen_tool` is a valid tool-scope `applies_to`, and you are right that no `schema_version` bump is needed since no new predicate is introduced. Severity and confidence sit exactly where the neighbors do (OAI-010 is low/0.65, ADK-009 low/0.7), and the pack id `autogen_observability` follows the `<category>_<topic>` convention. I also appreciate that the explanation is genuinely AutoGen framing rather than a copy of the OAI-010 text; the point about a print landing mid-transcript and reading as though it were part of the exchange is the right consequence to lead with. Picking AG2-018 to steer clear of the open PRs contesting AG2-013 through AG2-017 was thoughtful, and the ID is unclaimed by any other open PR. The engine half at trustabl/trustabl#153 mirrors this file byte for byte, so `rules-sync` stays green as long as the two land together. One housekeeping note: every shipped rule needs a rationale doc in trustabl-rulebook before release; we can handle that side or pair on it, it is not on you to write. Approving, and I will coordinate the paired merge.

PR #75, opened 2026-08-24
docs: complete the README layout tree for all ten packs

bradAGI:
The layout tree documented four of the ten packs and carried a note asking readers to browse the directory listing "until this section catches up." This catches it up: `claude_skill/`, `langchain/`, `crewai/`, `autogen/`, `pydantic_ai/`, and `vercel_ai/` are added and the note is dropped.

**Checking the existing entries against the packs turned up six that had gone stale** as rules shipped. None of these appeared in the tree:

| file | missing from the tree |
|---|---|
| `claude_sdk/agent_safety.yaml` | CSDK-121..124 (the `120/130/131` shorthand also hid the real set) |
| `claude_sdk/repo.yaml` | CSDK-204 |
| `claude_sdk/subagent_safety.yaml` | CSDK-112 |
| `claude_sdk/tool_definition.yaml` | CSDK-017, CSDK-018 |
| `openai_sdk/agent_safety.yaml` | OAI-112 |
| `mcp/tool_definition.yaml` | MCP-015..022 — the whole go/csharp/php/rust set |

Those are corrected here too. The MCP one seemed worth calling out: multi-language MCP coverage is a genuinely notable capability and the README gave no sign it existed.

Also added a line where the ID convention is stated, noting the two packs that don't follow it — `autogen/` numbers its agent rules AG2-001..006 in the tool range, and `claude_skill/` groups CSKILL-NNN by topic rather than scope. Both currently read as violations of the documented scheme.

**Verification.** Not checked by eye — I wrote a script that parses the tree out of the README, expands the `A-001..005` range notation, and diffs the claimed IDs against every `- id:` in every file, asserting both directions plus "no rule file left undocumented":

```
OK: 85 rule files documented, every rule ID matches the tree
$ trustabl rules validate .
OK: 85 rule pack(s), 206 rule(s) valid under rule schema version 14
```

Happy to contribute that script as a CI check if you'd want the tree kept honest automatically — it's ~40 lines of stdlib Python and would have caught all six drifts at the commit that introduced them. Say the word and I'll open a separate PR.

Deliberately documents only what's on `main` today, so it doesn't conflict with the rule PRs I have open.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,20 +50,20 @@ Rules are grouped by `<category>/<topic>.yaml`:
 ```
 manifest.yaml                         schema_version (metadata, not a rule)
 claude_sdk/                           Claude Agent SDK rules (CSDK-NNN)
-├── agent_safety.yaml                 CSDK-101..105 (agent); CSDK-120/130/131 (agent: permissionMode bypass + query() main-thread grants)
+├── agent_safety.yaml                 CSDK-101..105 (agent, python); CSDK-120..124, CSDK-130, CSDK-131 (agent, typescript)
 ├── code_execution.yaml               CSDK-107 (python), CSDK-011 (typescript)
 ├── error_handling.yaml               CSDK-005
 ├── idempotency.yaml                  CSDK-006 (python), CSDK-016 (typescript)
 ├── network.yaml                      CSDK-003
 ├── path_safety.yaml                  CSDK-004 (python), CSDK-012 (typescript fs-write)
-├── repo.yaml                         CSDK-201, CSDK-202 (repo scope, permission bypass)
+├── repo.yaml                         CSDK-201, CSDK-202 (permission bypass), CSDK-204 (max_turns) — repo scope
 ├── repo_hygiene.yaml                 CSDK-203 (repo scope, CLAUDE.md missing)
 ├── shell_safety.yaml                 CSDK-108 (python), CSDK-010 (typescript)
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI, this is careful work. I walked the whole tree against `origin/main` and everything you added or corrected checks out: the six drifts are real, the `MCP-015..022` multi-language set and `CSDK-121..124` were genuinely invisible, and nothing in the new tree names a rule that does not exist. Dropping the stale note is the right call.

Two things before I take it. `ADK-111` (MCPToolset with no `tool_filter`) landed in `google_adk/agent_safety.yaml` after you branched, and that line still reads `ADK-101..108, ADK-110`. A rebase and a re-run of your checker should pick it up.

The ID convention sentence is not quite right. It is not two packs: `CSDK-107` and `CSDK-108` are tool scope sitting in the `1NN` range, `CREW-108` likewise, and `ADK-008`, `VAI-006..009` and `VAI-012` sit in the tool range at agent and repo scope. That is fourteen rules across five packs, plus `claude_skill/`. I would rather the line said the numbering is advisory and that a number of rules predate it, than name two packs and leave the rest reading as compliant. While you are in there, please rewrap that paragraph; the inserted sentence runs well past the file's wrap width.

No engine-side pair is needed for this one, `scripts/check-rules-sync.sh` deliberately excludes docs from the comparison. And yes, please do open the checker as its own PR.

Happy to take this once `ADK-111` and the convention sentence are sorted.

sairenchristianbuerano (approved):

PR #65, opened 2026-08-24
feat(langchain): add LC-008, LC-009, LC-016 error and idempotency rules

Paramjeet-singh-neu:
## Summary
- Add **LC-008**: Python LangChain tool raises without a structured error contract (`has_raise` + no `has_try_except`), matching CSDK-005 / OAI-008 / MCP-006 / ADK-005.
- Add **LC-009**: Python mutating LangChain tool has no idempotency key (`name_has_prefix` + negated `param_name_matches`), matching CREW-006 / PYD-007 / CSDK-006.
- Add **LC-016**: TypeScript LangChain.js counterpart of LC-009 (camelCase prefixes + Zod schema param names), matching CSDK-016.

LangChain already discovers `@tool` / `StructuredTool` / `tool()` and ships shell/SSRF/code-exec rules, but it was missing the error-contract and mutating-tool idempotency checks that the stronger packs have. AgentExecutor / LangGraph `ToolNode` retries do not make these safe: LC-102's `max_iterations` bounds steps, not whether one mutating step is safe to retry or how the model interprets an opaque `ToolMessage`.

No new predicates and no `schema_version` bump — both matches are already used by other packs.

Paired PRs (same branch name `feat/langchain-error-idempotency`):
- Engine fixture + tests: opened next
- Rulebook rationale: opened next

## Test plan
- [ ] `trustabl rules validate .` on this pack
- [ ] Engine `go test ./internal/rules/ -run TestPolicyRules` (fire + silent cases for all three IDs)
- [ ] Confirm LC-007 / LC-015 (network timeout, open PR #52) do not collide — this pack uses 008/009/016

Made with [Cursor](https://cursor.com)

PR #77, opened 2026-08-24
feat(langchain): add LC-008, LC-016 — mutating tool has no idempotency key

eulestadt:
LangChain was the only one of seven packs without an idempotency rule, so this adds LC-008 for Python and LC-016 for TypeScript. Same `name_has_prefix` + `not param_name_matches` predicates as CSDK-006/016, OAI-009/019, ADK-006, CREW-006 and MCP-007, and the same medium/0.55 and medium/0.5 calibration — I'd rather not have LangChain drift from the family on a rule this well established. No new predicates, so no schema bump.

The explanation text is LangChain-specific rather than shared because the retry surfaces are. `AgentExecutor` retries a timed-out step, a LangGraph tool node re-enters on a retry policy that's configured nowhere near the tool definition, and the model can simply pick the same tool again after an ambiguous result. The middle one is what convinced me this is worth flagging statically — the tool author frequently can't see that retries are enabled from the file they're editing.

Worth saying up front that LC-102's iteration cap doesn't cover this, since that's the obvious "we already handle it" objection. It bounds how many times the loop runs; a capped loop that charged a card twice has still charged it twice.

I scanned a sample project rather than only running the unit tests. `create_order` fires; `create_refund`, which has the same mutating prefix but takes an `idempotency_key`, stays quiet; `get_order` stays quiet. The same three shapes behave the same way for LC-016 (`createCharge` / `createPayout` / `getBalance`). The guarded-but-mutating case is the one I cared about — a rule that fires on both is just noise.

Known misses, since they set the confidence: the prefix list isn't a verb ontology, so `submit_transfer` or `apply_discount` slip through, and a tool that accepts an `idempotency_key` and never forwards it downstream goes quiet while staying unsafe. That's why 0.55 and not higher.

Fixture mirror is [trustabl#123](https://github.com/trustabl/trustabl/pull/123), rationale doc is [trustabl-rulebook#43](https://github.com/trustabl/trustabl-rulebook/pull/43). This one needs to merge first — `rules-sync` looks up the paired branch with `git ls-remote` against this repo, so it can't see a branch on a fork, and the engine PR's check stays red until this lands.

Unrelated to these IDs, but flagging it since I just went through the same exercise on my MCP PR: #52 and #81 both claim `LC-007`. Whichever merges second will need a renumber, and a duplicate ID stops the loader at startup rather than failing a test.

### DIFF
--- a/langchain/idempotency.yaml
+++ b/langchain/idempotency.yaml
@@ -1,0 +1,95 @@
+policy:
+  id: langchain_idempotency
+  name: LangChain mutating-tool idempotency
+  category: langchain
+  description: >
+    Flags mutating LangChain tools (create/send/refund/...) without an
+    idempotency key. AgentExecutor and LangGraph re-invoke tool nodes under
+    timeouts and ambiguous failures, so without a key the same side effect
+    can fire twice.
+
+rules:
+  - id: LC-008
+    title: Mutating LangChain tool has no idempotency key
+    severity: medium
+    confidence: 0.55
+    language: python
+    applies_to:
+      - langchain_tool
+    scope: tool
+    match:
+      all:
+        - name_has_prefix:
+            - create_
+            - send_
+            - delete_
+            - post_
+            - update_
+            - refund_
+            - charge_
+            - issue_
+        - not:
+            param_name_matches:
+              contains:
+                - idempot
+              exact:
+                - request_id
+                - txn_id
+    explanation: >
+      Tool name suggests a side effect (create/send/refund/…). LangChain retries
+      a tool call when AgentExecutor or a LangGraph tool node times out, when
+      the model re-selects the same tool after an inconclusive result, or when
+      a response is lost after the side effect already committed
+      (at-least-once delivery). Without an idempotency key the same action can
+      fire twice — a duplicate charge, a double-sent message, a repeated
+      delete. An iteration cap such as max_iterations does not help: it bounds
+      how many steps the executor takes, not whether one mutating step may run
+      twice. The downstream service must also honor the key for this
+      protection to be effective.
+    fix: >
+      Add an `idempotency_key: str` parameter and pass it through to the backing
+      API so a retried call is recognized and deduplicated rather than
+      re-executed.
+
+  - id: LC-016
+    title: TypeScript LangChain mutating tool has no idempotency key
+    severity: medium
+    confidence: 0.5
+    language: typescript
+    applies_to:
+      - langchain_tool
+    scope: tool
+    match:
+      all:
+        - name_has_prefix:
+            - create
+            - send
+            - delete
+            - post
+            - update
+            - refund
+            - charge
+            - issue
+        - not:
+            param_name_matches:
+              contains:
+                - idempot
+              exact:
+                - requestId
+                - request_id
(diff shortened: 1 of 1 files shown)


eulestadt:
Companion PRs:

- Engine fixture + tests: https://github.com/trustabl/trustabl/pull/123
- Rulebook rationale: https://github.com/trustabl/trustabl-rulebook/pull/43

eulestadt:
ID collision note, so reviewers can pick:

- **#55** (Brad) already uses **LC-008** for Python error-handling. That PR is the right home for `has_raise`; we should not fight it.
- **#62** (Brad) is Python-only idempotency as **LC-017**, with no engine fixture PR and no rulebook PR.
- **#65** (Param) bundles error + idempotency (LC-008 / LC-009 / LC-016).

This PR is the **Python + TypeScript pair** (LC-008 / LC-016) matching CSDK-006/016 and this pack’s existing 00N-Python / 01N-TypeScript split (LC-005/013, LC-006/014), with fixture tests ([trustabl#123](https://github.com/trustabl/trustabl/pull/123)) and a full rationale ([rulebook#43](https://github.com/trustabl/trustabl-rulebook/pull/43)).

Happy to retarget IDs if that’s easier to merge — the TypeScript half is the piece #62 does not cover.

jhumel-code:
Thanks @eulestadt, this is a model rules PR. `LC-008` and `LC-016` mirror the established idempotency family faithfully (I diffed the match blocks against `CSDK-006` and `CSDK-016` on main: same prefixes, same guard params, same medium 0.55 and medium 0.5 calibration), both predicates exist in the engine on main, `langchain_tool` is valid at tool scope, and you are right that no schema bump is needed. The explanation text is genuinely LangChain framed rather than copied, and the paragraph preempting the "LC-102 already covers this" objection is exactly the kind of reasoning I want living in these packs. The fixture mirror in trustabl/trustabl#123 is byte identical with real fire and silent cases, and the rationale doc in trustabl-rulebook#43 covers the grounding requirement; that doc needs to land before these ship, and we are happy to handle or pair on that part.

One coordination item before merge: `LC-008` is also claimed by open PRs trustabl/trustabl#122, trustabl/trustabl#132, trustabl/trustabl-rules#55, and trustabl/trustabl-rules#65, and `LC-016` by trustabl/trustabl#122 and trustabl/trustabl-rules#65. Whichever lands first keeps the IDs and the others renumber; I will sequence that on our side, and we will land this and trustabl/trustabl#123 back to back so `rules-sync` is only red for the gap between them. Thanks also for flagging the `LC-007` contention.

eulestadt:
Thanks. Glad the match blocks lined up. Sounds good; I'll stay out of the ID sequencing unless you need something from me.

sairenchristianbuerano:
**Closing.** `LC-008` goes to trustabl/agent-reliability-rules#55, which uses it for the error-contract rule. The LangChain idempotency rule you wrote here ships as `LC-017` via trustabl/agent-reliability-rules#62 — the one version of that rule that collides with no other id. `LC-016` is not carried forward.

For the record, your earlier renumber of `MCP-023` to `MCP-029` was the right instinct; this is the same problem at a larger scale.

PR #78, opened 2026-08-24
feat(mcp): add MCP-201, project ships no agent-guidance doc

bradAGI:
**MCP was the only pack with no repo-scope rule at all.** The `mcp` `applies_to` token is valid in the loader's `appliesToByScope` but no shipped rule used it — I found this auditing which engine capabilities have zero rule coverage. Every other pack ships the agent-guidance check: CSDK-203, OAI-202, ADK-201, LC-201, CREW-201, AG2-201, PYD-201, VAI-012.

The framing is MCP's own rather than ADK-201 with the names swapped. **Everything this repo exports crosses a trust boundary** to whatever client and model connect to it, and the tool name, description, and input schema are the entire contract those consumers see. So the consequence is unusually indirect: a tool added without the local conventions doesn't fail *here* — it degrades tool selection in every downstream client, and the server's authors are not the ones who observe it. That's the argument for writing the conventions down in-tree.

The fix asks for what an MCP `AGENTS.md` specifically needs: the namespace prefix, what a description must cover for a client with no other context about this server, how input schemas are typed, and which operations are destructive and what gates them.

```
$ trustabl rules validate .
```

Fire (FastMCP server, no AGENTS.md or CLAUDE.md): `MCP-201`
Silent (same server with an AGENTS.md at the root): no findings

### DIFF
--- a/mcp/repo_hygiene.yaml
+++ b/mcp/repo_hygiene.yaml
@@ -1,0 +1,45 @@
+policy:
+  id: mcp_repo_hygiene
+  name: MCP server repo hygiene
+  category: mcp
+  description: >
+    Repo-scoped hygiene rules for projects that implement an MCP server. Fire
+    once per scan against the repo manifest and inventory rather than per tool.
+
+rules:
+  - id: MCP-201
+    title: MCP server project ships no agent-guidance doc (AGENTS.md/CLAUDE.md)
+    severity: low
+    confidence: 0.9
+    applies_to:
+      - mcp
+    scope: repo
+    match:
+      all:
+        - repo_has_sdk_in_code:
+            - mcp
+        - not:
+            repo_component_present:
+              - agents_md
+              - claude_md
+    explanation: >
+      The project implements an MCP server in code but ships no agent-guidance
+      doc (AGENTS.md or CLAUDE.md) at the repo root. AGENTS.md is the
+      cross-vendor convention an editing coding agent reads before it acts, so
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI, this is a well-found gap. Every other SDK pack ships a `repo_hygiene.yaml` and `mcp/` had nothing at repo scope, so the audit that turned this up was the right instinct. `MCP-201` is free on `main` (the `mcp` IDs stop at `MCP-022`) and no other open PR claims it.

The mechanics are all correct where this rule is easy to get wrong. `mcp` is a valid `applies_to` at repo scope, `repo_has_sdk_in_code: mcp` uses the SDK enum token rather than the category token (the loader rejects the latter, and it is the usual trip hazard here), severity and confidence match `LC-201` and `CSDK-203`, and omitting `language:` is right: repo scope does not default it, so this also covers the Go, Rust, C#, and PHP MCP servers, not just Python. No new predicates, so no `schema_version` bump is the correct call.

The explanation and fix are genuinely MCP-framed rather than `LC-201` with the names swapped, which is what I want from a new pack rule.

One thing before it ships: each rule needs a rationale doc in `trustabl-rulebook`, and there is no MCP repo-hygiene doc for this to slot into yet. That is on the maintainer side, and I am happy to pair on it.

This lands together with trustabl/trustabl#154, where the fixture copy is byte-identical, so `rules-sync` stays green.

sairenchristianbuerano (approved):

PR #82, opened 2026-08-24
feat(mcp): add MCP-023, tool prints to stdout, corrupting a stdio transport

bradAGI:
MCP had no observability rule. OpenAI ships OAI-010 and ADK ships ADK-009 for the same `print()` pattern — both at `low` severity, framed as a lost-diagnostic problem.

**It isn't a lost-diagnostic problem here, which is why this ships at `medium` instead.** On a stdio server — the FastMCP default, and the usual way an editor or desktop client launches a server — stdout *is* the protocol channel. The client reads newline-delimited JSON-RPC messages from it, so a loose `print` interleaves with those frames: the client hits a parse error on a line that isn't JSON and, depending on the client, drops the response, logs a protocol error, or tears down the connection.

The reason it's worth flagging rather than leaving to review is that **the failure doesn't look like its cause**. The symptom is a tool call that returns nothing, or a server that disconnects mid-session; the `print` that broke it reads like harmless debugging and sits nowhere near the error.

OAI-010's explanation already anticipates exactly this — it warns that a tool "ever exposed over an MCP stdio server" will corrupt the stream. This rule is that case as the primary reading rather than the footnote.

Confidence is 0.7 rather than higher because a server run over HTTP or SSE escapes the corruption — though it still loses the diagnostic, since the model only ever sees the return value. The fix names **stderr** specifically, which the transport leaves alone and clients typically capture as server logs, plus the caveat about nothing in the process reconfiguring logging onto stdout.

```
$ trustabl rules validate .
```

Fire (`print(f"looking up {order_id}")` in a FastMCP tool handler): `MCP-023`
Silent (`logger.info(...)` with `basicConfig(stream=sys.stderr)`): no findings

### DIFF
--- a/mcp/observability.yaml
+++ b/mcp/observability.yaml
@@ -1,0 +1,41 @@
+policy:
+  id: mcp_observability
+  name: MCP tool observability hygiene
+  category: mcp
+  description: >
+    Rules covering how an MCP tool handler emits diagnostics. On a stdio
+    server — the default transport — stdout is the protocol channel, so a
+    handler that prints to it is not writing a log line, it is writing into the
+    JSON-RPC frame stream.
+
+rules:
+  - id: MCP-023
+    title: MCP tool prints to stdout, corrupting a stdio transport
+    severity: medium
+    confidence: 0.7
+    language: python
+    applies_to:
+      - mcp_tool
+    scope: tool
+    match:
+      has_print_call: true
+    explanation: >
+      The tool handler calls print(), which writes to the process's stdout. On
+      an MCP server using the stdio transport — the default for FastMCP and the
+      usual way an editor or desktop client launches a server — stdout is not a
+      log sink, it is the protocol channel: the client reads newline-delimited
+      JSON-RPC messages from it. A loose print interleaves with those frames, so
+      the client hits a parse error on a line that is not JSON and, depending on
+      the client, drops the response, logs a protocol error, or tears down the
+      connection. The failure does not look like the cause: the symptom is a
+      tool call that returns nothing or a server that disconnects mid-session,
+      while the print that broke it reads like harmless debugging. A server run
+      over HTTP or SSE instead escapes the corruption, but the diagnostic is
+      still lost — the model never sees stdout, only the return value.
+    fix: >
+      Remove the print(). Diagnostics from an MCP server belong on stderr, which
+      the transport leaves alone and clients typically capture as server logs —
+      emit through a module logger (logging.getLogger(__name__)) configured with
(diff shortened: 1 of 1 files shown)


eulestadt:
Heads-up on an ID overlap, and it's already resolved on my side — no action needed from you.

This PR and my #80 both shipped as `MCP-023` about four minutes apart. Since a duplicated rule ID is a loader refusal at startup rather than a test failure, two green PRs merging an hour apart would have broken every scan until someone renumbered.

Your stack claims 023 through 028 (#82, #85, #88, #91), so yielding was one edit for me versus five for you. **I've renumbered mine to `MCP-029`** across the rules, fixture and rulebook PRs. `MCP-023` is yours; nothing here needs to change.

While I was checking the ranges: **#52 and #81 both claim `LC-007`** — the network-timeout pair and the error-contract rule. Whichever lands second will need the same treatment. Mentioning it only because I'd just done the exercise.

Separately, #55's note that `PredHasRaise` can't support a TS error-contract rule was correct and I've opened [trustabl#169](https://github.com/trustabl/trustabl/pull/169) to close it — `tsHandlerFacts` now records `throws` / `try_catch` and both predicates branch on `IsTSOrJS` the way `PredHasShellCall` already did. It's engine-only and merges on its own, so TS counterparts to CSDK-005 / OAI-008 / ADK-005 / MCP-006 become writable after it. All yours if you want them.

jhumel-code:
Thanks @bradAGI, and thanks for the write up. The severity argument holds. On a stdio server stdout is the framing channel, so this is a correctness and availability bug rather than the lost diagnostic that `OAI-010` and `ADK-009` describe, and `medium` at `0.7` sits comfortably beside `MCP-007` (medium, 0.55) and `MCP-002` (medium, 0.85) in the same pack.

The rest checks out against `main`. `MCP-023` is unused there and no other open PR claims it, `mcp_tool` is valid at `tool` scope, `has_print_call` already exists so you are right that no `schema_version` bump belongs here, and every required field is present with nothing the closed schema would reject. `language: python` is the right call, since the predicate is Python AST backed and `MCP-015..022` cover the other languages. The fixture copy in trustabl/trustabl#156 matches this file exactly.

Two gaps worth capturing in the rationale doc rather than changing here: the rule only sees the tool function body, so a print in a shared helper or a lifespan hook corrupts the same stream unseen, and `sys.stdout.write` slips past the bare callee match. Optionally, add a `mcp/observability.yaml` line to the `mcp/` block in the README layout tree.

`MCP-023` needs a rationale doc in `trustabl-rulebook` before it ships; that is on us and I am happy to pair on it. This must land together with trustabl/trustabl#156.

PR #84, opened 2026-08-24
feat(claude_sdk): add CSDK-020, TypeScript tool HTTP call has no timeout

bradAGI:
CSDK-003 covers the Python side of network timeouts; the TypeScript half was missing, even though this pack ships TS rules throughout (CSDK-010..014, CSDK-120..131). OpenAI (OAI-016, OAI-024) and the Vercel AI SDK (VAI-011) already use `has_http_call_without_timeout` for exactly this.

The consequence framing is what makes it worth stating rather than porting VAI-011's text: **an unresponsive host stalls the conversation rather than failing it.** The model gets no result and no error, so the turn can't advance — and a `max_turns` cap doesn't help, because the run is stuck *inside* a single turn rather than taking too many. In a server embedding the SDK it also holds the request worker for the duration.

Compounds with CSDK-013 the way VAI-011 compounds with VAI-003: a tool that fetches a model-controlled URL and can't time out can be pointed at an internal host that simply never answers.

```
$ trustabl rules validate .
```

Fire (bare `await fetch(...)` in a `tool(...)` handler): `CSDK-013, CSDK-020, CSDK-203`
Silent (`signal: AbortSignal.timeout(15_000)`): `CSDK-013, CSDK-203`

(CSDK-013 is the pre-existing SSRF rule firing on the fixture's template-string URL — which is also the compounding case the explanation describes. CSDK-203 is the missing-CLAUDE.md repo rule.)

### DIFF
--- a/claude_sdk/network.yaml
+++ b/claude_sdk/network.yaml
@@ -53,3 +53,34 @@ rules:
     fix: >
       Pass `timeout=` (typically 5–30s) to the request. Surface failures as a
       structured error the model can react to.
+
+  - id: CSDK-020
+    title: TypeScript Claude SDK tool HTTP call has no timeout
+    severity: high
+    confidence: 0.6
+    language: typescript
+    applies_to:
+      - claude_sdk_tool
+    scope: tool
+    match:
+      has_http_call_without_timeout: true
+    explanation: >
+      This TypeScript Claude SDK tool makes an outbound HTTP call (fetch /
+      axios / got / undici) with no deadline — no signal, timeout, or
+      abortSignal option. Node's fetch has no implicit deadline, so a slow or
+      unresponsive host blocks the tool callback until the socket eventually
+      dies. Because the call sits inside the agent's turn, that stalls the
+      conversation rather than failing it: the model gets no result and no
+      error, the turn cannot advance, and a max_turns cap does not help because
+      the run is stuck inside a single turn rather than taking too many. In a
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI. `CSDK-020` is free on `main` and no other open PR claims it, `has_http_call_without_timeout` already exists in the engine so you are right that no `schema_version` bump belongs here, and `severity: high` with `confidence: 0.6` matches OAI-016 and VAI-011 exactly, which is where I would put it given the predicate cannot see a signal defined on another line and passed by identifier. `language: typescript` and `applies_to: [claude_sdk_tool]` are both correct for `scope: tool`, and the CSDK-013 cross reference is accurate.

Two nits before this goes in. The explanation says `max_turns`, but this is a TypeScript rule and the TS SDK spells that option `maxTurns`; the snake_case form reads as the Python option and will look wrong to the audience for this rule. And the `Layout` section of the README still lists `network.yaml` as `CSDK-003` only, so please make it `CSDK-003 (python), CSDK-020 (typescript)` the way `code_execution.yaml` and `ssrf.yaml` are already listed.

This also needs a rationale doc in `trustabl-rulebook`, ideally naming the shorthand `{ signal }` and spread blind spots; that part is on us and I am glad to pair on it. Merging in lockstep with trustabl/trustabl#158, so the prose fix needs mirroring into the fixture there too. Happy to take it once those two items are fixed.

PR #86, opened 2026-08-24
feat(google_adk): add ADK-114, TypeScript FunctionTool HTTP call has no timeout

bradAGI:
ADK-003 covers the Python side of network timeouts; the TypeScript half was missing, even though the pack ships TS rules (ADK-013, ADK-015, ADK-016, ADK-109). OpenAI (OAI-016, OAI-024) and the Vercel AI SDK (VAI-011) already use `has_http_call_without_timeout` for exactly this.

**The composition ADK encourages makes the stall worse rather than better**, which is the part worth having in the finding text:

- inside a `SequentialAgent`, a stalled tool blocks every step after it;
- inside a `ParallelAgent`, the whole fan-out waits on its slowest branch — so one unresponsive host stalls work that has nothing to do with it.

And the limit people reach for doesn't apply: **ADK-108's `max_iterations` bounds how many times a `LoopAgent` goes round, not how long one tool call may run.** No configured limit breaks this stall.

Compounds with ADK-016 the same way VAI-011 compounds with VAI-003 — a tool that fetches a caller-controlled URL and can't time out can be steered at an internal host that never answers.

Numbered ADK-114 to leave room for ADK-111 in the open PR #51 and ADK-112 in my #79.

```
$ trustabl rules validate .
```

Fire (bare `await fetch(...)` in a `new FunctionTool({ ..., execute })` handler): `ADK-016, ADK-114, ADK-201`
Silent (`signal: AbortSignal.timeout(15_000)`): `ADK-016, ADK-201`

(ADK-016 is the pre-existing TS SSRF rule firing on the template-string URL — also the compounding case described. ADK-201 is the missing-AGENTS.md repo rule.)

My first fixture used `new FunctionTool(fn, {...})` and nothing fired, including the pre-existing TS rules — the adk-js shape is the options-object form with an `execute` callback. Worth flagging since the Python `FunctionTool(fn)` wrapper reads as the obvious analogue and isn't what the TS SDK does.

### DIFF
--- a/google_adk/network.yaml
+++ b/google_adk/network.yaml
@@ -51,3 +51,37 @@ rules:
       Pass `timeout=` (typically 5–30 seconds) to the request. Choose a
       value that's tight enough to fail fast and loose enough to allow
       legitimate slow responses for this endpoint.
+
+  - id: ADK-114
+    title: TypeScript FunctionTool HTTP call has no timeout
+    severity: high
+    confidence: 0.6
+    language: typescript
+    applies_to:
+      - adk_function_tool
+    scope: tool
+    match:
+      has_http_call_without_timeout: true
+    explanation: >
+      This TypeScript ADK FunctionTool makes an outbound HTTP call (fetch /
+      axios / got / undici) with no deadline — no signal, timeout, or
+      abortSignal option. Node's fetch has no implicit deadline, so a slow or
+      unresponsive host blocks the tool callback until the socket eventually
+      dies. What that costs depends on where the tool sits in the agent tree,
+      and the composition ADK encourages makes it worse rather than better: a
+      stalled tool inside a SequentialAgent blocks every step after it, and
+      inside a ParallelAgent the whole fan-out waits on its slowest branch, so
+      one unresponsive host stalls work that has nothing to do with it.
+      ADK-108's max_iterations bounds how many times a LoopAgent goes round, not
+      how long one tool call may run, so no configured limit breaks the stall.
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI, this is a clean pair. I checked ADK-114 against `main`: `has_http_call_without_timeout` exists and is TypeScript-only (it reads the `http_no_timeout` fact set by `tsHandlerFacts`), `adk_function_tool` is a valid tool-scope `applies_to` value, `language: typescript` is stated explicitly, and `high` / `0.6` matches the two shipped rules built on the same predicate, OAI-016 and VAI-011. No new predicate, so leaving `schema_version` at 14 is the right call.

The explanation is the strongest part. The SequentialAgent and ParallelAgent framing and the ADK-016 compounding note are genuinely ADK-specific rather than OAI-016 text with the names swapped. One small thing: ADK-108 is a Python-only rule, so a TypeScript reader who follows that reference lands on something that cannot fire in their repo. Describing it as the LoopAgent iteration cap rather than by rule ID would travel better, though I would not hold the PR for it.

Two coordination notes. ADK-114 is uniquely claimed across the open PRs, so no collision there, but ADK-111 has since landed on `main`, so trustabl/trustabl-rules#51 will need to renumber. And this rule needs a rationale doc in `trustabl-rulebook` before it ships; we can take that or pair on it, it is not on you. The engine half trustabl/trustabl#160 carries byte-identical YAML, so the two should land together.

PR #87, opened 2026-08-24
feat(openai_sdk): add OAI-025, OAI-026 tool description quality rules

bradAGI:
Ports the CSDK-017/018 description-quality pair (merged in #40) to the OpenAI Agents SDK. OAI-001 only checks that a docstring *exists*, so a tool whose docstring reads `"""TODO: describe this tool."""` or `"""Gets data."""` passes today while giving the model no selection signal.

Two OpenAI-specific points, both of which are things people reasonably assume are already covered:

- **Strict schemas don't compensate.** OAI-003's `strict_mode` constrains the *shape of the arguments*, never whether calling this tool was the right move. A placeholder description produces a well-formed call to the wrong tool — schema validation passes and the run is still wrong.
- **With handoffs, the blast radius is the whole conversation.** An agent picks between its own tools and its peers' on the strength of these strings, so a thin description doesn't just cause a wrong call — it can route the conversation to the wrong agent.

Mis-selection also spends a turn against the OAI-112 `max_turns` budget, and nothing in a trace attributes the wasted turn to the description that caused it.

OAI-025's fix names `description_override` on `@function_tool`, which is the SDK-idiomatic place to put it when the function name can't carry the meaning.

```
$ trustabl rules validate .
```

Fire (one tool with `"""TODO: describe this tool."""`, one with `"""Gets data."""`): `OAI-004, OAI-025, OAI-026, OAI-201, OAI-202`
Silent (full description naming when to prefer the neighboring tool): `OAI-004, OAI-201, OAI-202`

(OAI-004, OAI-201, OAI-202 are pre-existing rules firing on the fixture's decorator config, default tracing, and missing CLAUDE.md.)

OAI-026 pairs `description_length_lt: 40` with `has_docstring: true` so it doesn't double-report against OAI-001 on an empty docstring, same as CSDK-018.

### DIFF
--- a/openai_sdk/tool_definition.yaml
+++ b/openai_sdk/tool_definition.yaml
@@ -95,3 +95,65 @@ rules:
       Provide a concise `description` string in the `tool({...})` options stating
       what the tool does and when the model should call it. The description is the
       model's primary routing signal alongside the tool name.
+
+  - id: OAI-025
+    title: Tool description is a placeholder
+    severity: low
+    confidence: 0.85
+    language: python
+    applies_to:
+      - openai_tool
+    scope: tool
+    match:
+      has_description_text:
+        - todo
+        - tbd
+        - fixme
+        - placeholder
+        - no description
+        - does stuff
+    explanation: >
+      The docstring passes the OAI-001 has-a-description check but carries a
+      placeholder marker instead of real content. The Agents SDK forwards this
+      text to the model verbatim as the tool's selection signal, so a stub like
+      "TODO: describe this tool" is functionally indistinguishable from no
+      description at all and the model still has to guess from the function
+      name. Strict schemas do not compensate: OAI-003's strict_mode constrains
+      the shape of the arguments, never whether calling this tool was the right
+      move, so a placeholder description produces a well-formed call to the
+      wrong tool. Mis-selection also spends a turn against the OAI-112 max_turns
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI. This is well put together. Both rules carry every required field, `openai_tool` is a valid `applies_to` for `scope: tool`, `language: python` is stated explicitly, and the severity and confidence values match the CSDK-017/018 pair they are ported from, which is the right precedent to anchor to. `has_description_text` and `description_length_lt` both exist in the engine on `main` under schema 14, so no bump was needed and you were right to leave `manifest.yaml` alone. The OAI-003 and OAI-112 cross-references in the explanation text are accurate, and the framing reads as genuinely OpenAI specific rather than Claude text with the names swapped.

No other open PR claims OAI-025 or OAI-026, so there is no renumbering risk on these IDs.

One thing to be aware of rather than to change: a placeholder docstring shorter than 40 characters trips both rules on the same tool (`"""TODO: describe this tool."""` is 26 characters), and the `placeholder` needle will also hit a tool that legitimately does template substitution. Both behaviors are inherited from CSDK-017/018, so I would rather keep the pair consistent than diverge here.

The engine half trustabl/trustabl#164 mirrors this file exactly, so the two need to land together. New rules also need a rulebook rationale doc before they ship, which we can handle or pair on.

PR #88, opened 2026-08-24
feat(mcp): add MCP-025, MCP-026 tool description quality rules

bradAGI:
Ports the CSDK-017/018 description-quality pair (merged in #40) to MCP. MCP-001 only checks that a docstring *exists*, so a tool whose docstring reads `"""TODO: describe this tool."""` or `"""Gets data."""` passes today while giving connecting clients no selection signal.

**A placeholder costs more across a protocol boundary than inside an in-process SDK**, for three reasons the rule text makes explicit:

- **The consumer has no fallback context.** It can't read the source, the neighboring tools, or the project's docs the way a developer working in the repo can. The `tools/list` string is genuinely all it gets.
- **The server's authors never see the consequence.** Mis-selection surfaces in someone else's client session as a wrong answer — not as an error on this side of the connection. Nothing brings it back to the maintainer.
- **The competition isn't just this server's tools.** A client typically connects several servers at once and the model chooses across all of them from these strings alone, so a thin description loses to a well-described tool from an unrelated server that only approximately fits the request.

That third point drives MCP-026's fix: name the domain the tool acts on rather than assuming the server name conveys it, because the model sees the tool alongside every other connected server's.

```
$ trustabl rules validate .
```

Fire (one tool with `"""TODO: describe this tool."""`, one with `"""Gets data."""`): `MCP-025, MCP-026`
Silent (full description naming when to prefer the neighboring tool): no findings

MCP-026 pairs `description_length_lt: 40` with `has_docstring: true` so it doesn't double-report against MCP-001 on an empty docstring, same as CSDK-018.

### DIFF
--- a/mcp/tool_definition.yaml
+++ b/mcp/tool_definition.yaml
@@ -309,3 +309,68 @@ rules:
     fix: >
       Rename the method (or set the `#[tool]` `name = "..."` argument) to a
       verb-object form, e.g. `summarize_invoice`, `fetch_weather`.
+
+  - id: MCP-025
+    title: MCP tool description is a placeholder
+    severity: low
+    confidence: 0.85
+    language: python
+    applies_to:
+      - mcp_tool
+    scope: tool
+    match:
+      has_description_text:
+        - todo
+        - tbd
+        - fixme
+        - placeholder
+        - no description
+        - does stuff
+    explanation: >
+      The docstring passes the MCP-001 has-a-description check but carries a
+      placeholder marker instead of real content. The server advertises this
+      string across the protocol boundary in its tools/list response, so it is
+      the entire account of the tool that every connecting client and model
+      receives — and a stub like "TODO: describe this tool" is functionally
+      indistinguishable from no description at all. A placeholder is worse here
+      than in an in-process SDK for two reasons. The consumer has no other
+      context to fall back on: it cannot read the source, the neighboring tools,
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI. The port is faithful to CSDK-017/018 (same severity, same confidence, same predicates) and, more importantly, the explanation and fix text is genuinely rewritten for the protocol boundary instead of copied across with the wrong framing. That is exactly what I want in an SDK-specific pack. `MCP-025` and `MCP-026` are free IDs (`MCP-023` and `MCP-024` are claimed by other open PRs, nothing here collides), `mcp_tool` is correct for tool scope, and you are right that no `schema_version` bump is needed, since `has_description_text`, `has_docstring` and `description_length_lt` all exist on main at schema 14.

I read the fixture half in trustabl/trustabl#165 and it is byte identical to this file, so `rules-sync` will be satisfied as long as the two land together. Neither half should merge alone.

One note rather than a change request: `has_description_text` matches case-insensitively on any substring, so a task-management server whose tool reads "Add a todo item to the list." will fire `MCP-025`. I would rather keep the needle list identical to CSDK-017 than have the two diverge, so let us record that false positive in the rulebook rationale doc, which these two rules still need before they ship. Happy to pair on that part.

If you want a follow-up afterwards, the TypeScript side (the MCP-011 neighbourhood) has no placeholder or length rule yet.

PR #89, opened 2026-08-24
feat(google_adk): add ADK-115, ADK-116 tool description quality rules

bradAGI:
Ports the CSDK-017/018 description-quality pair (merged in #40) to the Google ADK. ADK-001 only checks that a docstring *exists*, so a tool whose docstring reads `"""TODO: describe this tool."""` or `"""Gets data."""` passes today while giving the model no selection signal.

Two ADK-specific points:

- **The docstring does double duty.** The ADK parses it for the per-parameter descriptions that accompany the generated schema, so a placeholder or one-clause docstring strips argument-level guidance at the same time as tool-level guidance. Both fixes ask for documented parameters, not just a longer summary.
- **In an agent tree, mis-selection doesn't stay local.** A tool picked wrongly inside one branch produces output the following agents treat as established fact — the error gets laundered into shared session state rather than surfacing where it began.

ADK-116 also names the mitigation people would reach for and why it isn't one: **ADK-102 and ADK-107's `before_tool_callback` can block a call it recognizes as wrong, but it can't supply the judgment the description was meant to provide.** A callback doesn't compensate for a thin description.

Numbered ADK-115/116 to leave room for ADK-111 in the open PR #51 and the IDs used by my other open PRs (#79, #86).

```
$ trustabl rules validate .
```

Fire (one `FunctionTool`-wrapped function with `"""TODO: describe this tool."""`, one with `"""Gets data."""`): `ADK-115, ADK-116, ADK-201`
Silent (full docstring with an `Args:` section): `ADK-201`

ADK-116 pairs `description_length_lt: 40` with `has_docstring: true` so it doesn't double-report against ADK-001 on an empty docstring, same as CSDK-018.

### DIFF
--- a/google_adk/tool_definition.yaml
+++ b/google_adk/tool_definition.yaml
@@ -127,3 +127,67 @@ rules:
       Pass a one-sentence `description` in the `new FunctionTool({...})` options
       that names what the tool does, the inputs it expects, and what it returns.
       Write it for the model, not for a human reader.
+
+  - id: ADK-115
+    title: FunctionTool description is a placeholder
+    severity: low
+    confidence: 0.85
+    language: python
+    applies_to:
+      - adk_function_tool
+    scope: tool
+    match:
+      has_description_text:
+        - todo
+        - tbd
+        - fixme
+        - placeholder
+        - no description
+        - does stuff
+    explanation: >
+      The docstring passes the ADK-001 has-a-docstring check but carries a
+      placeholder marker instead of real content. The ADK builds the tool
+      declaration it sends to the model from this docstring, so a stub like
+      "TODO: describe this tool" is functionally indistinguishable from no
+      docstring at all and the model still has to guess from the function name.
+      The ADK also parses the docstring for the per-parameter descriptions that
+      accompany the generated schema, so a placeholder body strips the
+      argument-level guidance at the same time as the tool-level guidance. In an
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI, this is a clean port. Both predicates already exist on `main`, so you are right that no `schema_version` bump belongs here, and `adk_function_tool` is a valid `applies_to` at `tool` scope. Severity and confidence match CSDK-017/018 and ADK-001 in the same pack, and neither ID is claimed by another open PR.

What I appreciated is that you did not just copy the Claude framing across. The docstring doing double duty for the per-parameter descriptions, and the point that `before_tool_callback` can block a wrong call but cannot supply the judgment a thin description was meant to carry, are both ADK-specific and both correct.

Two non-blocking observations. A placeholder docstring under 40 characters will report under `ADK-115` and `ADK-116` both, which is exactly what CSDK-017/018 do today, so I am fine keeping parity. And the `todo` needle will match a genuine docstring for a todo-list tool, also inherited, and low severity at 0.85 absorbs it.

Before this ships it needs a rationale doc in `trustabl-rulebook` under `google_adk`. That part is ours, not a demand on you, and I am happy to pair on it. I will land this together with the engine half, trustabl/trustabl#166, so `rules-sync` stays green.

PR #91, opened 2026-08-24
feat(mcp): add MCP-027, MCP-028 TypeScript description quality rules

bradAGI:
The TypeScript counterpart to MCP-025/026 (#88). MCP-011 only checks that a description *exists*, so a registration described `"TODO: describe this tool."` or `"Gets data."` passes today while publishing no selection signal to connecting clients.

Carries the same protocol-boundary argument as the Python pair, which is what makes this more than a lint:

- **The consumer has no fallback context** — it can't read this repo's source, the neighboring tools, or the project's docs.
- **The server's authors never observe the mis-selection** — it surfaces in someone else's client session as a wrong answer, not as an error on this side.
- **The competition isn't just this server's tools** — a client typically connects several at once and the model picks across all of them from these strings alone.

Plus the TypeScript-specific point in MCP-028: **the Zod input schema doesn't close the gap.** It constrains the arguments once the model has chosen this tool, not whether choosing it was right.

```
$ trustabl rules validate .
```

Fire (one `server.tool` described `"TODO: describe this tool."`, one `"Gets data."`): `MCP-027, MCP-028`
Silent (full description naming when to prefer the neighboring tool): no findings

MCP-028 pairs `description_length_lt: 40` with `has_docstring: true` so it doesn't double-report against MCP-011 on an empty description, same as CSDK-018.

**Merge note:** this and #88 both append to `mcp/tool_definition.yaml`, so whichever lands second will need a trivial rebase — the two rule blocks are independent and the resolution is "keep both." Ping me and I'll rebase, or take them in either order and I'll follow up. This is the only pair among my open PRs that touches the same file; the rest are all distinct files.

### DIFF
--- a/mcp/tool_definition.yaml
+++ b/mcp/tool_definition.yaml
@@ -309,3 +309,67 @@ rules:
     fix: >
       Rename the method (or set the `#[tool]` `name = "..."` argument) to a
       verb-object form, e.g. `summarize_invoice`, `fetch_weather`.
+
+  - id: MCP-027
+    title: TypeScript MCP tool description is a placeholder
+    severity: low
+    confidence: 0.85
+    language: typescript
+    applies_to:
+      - mcp_tool
+    scope: tool
+    match:
+      has_description_text:
+        - todo
+        - tbd
+        - fixme
+        - placeholder
+        - no description
+        - does stuff
+    explanation: >
+      The tool registration sets a description, so it passes MCP-011, but the
+      string is a placeholder rather than real content. The server publishes it
+      across the protocol boundary in its tools/list response, so it is the
+      entire account of the tool that every connecting client and model
+      receives, and "TODO: describe this tool" tells them nothing the tool name
+      did not. The consumer has no fallback: it cannot read this repo's source,
+      the neighboring tools, or the project's docs. The server's authors do not
+      see the result either, because mis-selection surfaces in someone else's
+      client session as a wrong answer rather than as an error on this side of
+      the connection.
+    fix: >
+      Replace the placeholder with a real description covering what the tool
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI, this is a clean pack change. Both `has_description_text` and `description_length_lt` already exist on `main`, so no `schema_version` bump is right, and pairing `description_length_lt: 40` with `has_docstring: true` is exactly the guard the schema comment asks for, keeping an empty description as MCP-011's finding. Severity, confidence and the needle list line up with the CSDK-017/018 precedent, and the explanation and fix are genuinely rewritten for the MCP protocol boundary rather than carried over from the Claude framing. The engine half in trustabl/trustabl#168 mirrors this byte for byte, which I checked.

The one blocker is not yours. MCP-027 and MCP-028 are also claimed by trustabl/trustabl-rules#100 (with trustabl/trustabl#135) for unrelated rules covering idempotency keys and ambiguous names. Whichever pair lands first keeps the numbers and the other renumbers and rebases, so I will sequence those rather than ask you to change anything yet.

One thing to carry forward: `todo` and `placeholder` as bare substrings will fire on a legitimate description such as "Create a new todo item", so that belongs in the known gaps section of the rulebook rationale doc this pack still needs before the rules ship. We can write that or pair with you on it, it is not a request on this PR.

sairenchristianbuerano:
**Selected for merge.** This pair keeps `MCP-027` and `MCP-028`.

Closing trustabl/agent-reliability-rules#100 + trustabl/agent-reliability-analyzer#135, which claimed both ids for different rules (TypeScript idempotency and ambiguous tool name).

PR #93, opened 2026-08-24
feat(langchain): add LC-021, LC-022 TypeScript description quality rules

bradAGI:
The TypeScript counterpart to LC-018/019 (#66). LC-010 only checks that a description *exists*, so a tool described `"TODO: describe this tool."` or `"Gets data."` passes today while giving the model no selection signal.

Three points, in the order they bite:

- **No docstring fallback.** LangChain.js has nothing behind the `description` field, so it's the entire account of the tool the model sees — unlike the Python side, where a thin description at least sits next to readable source.
- **The Zod schema doesn't compensate.** It constrains the shape of the arguments once the model has chosen this tool, never whether choosing it was right.
- **Mis-selection scales badly and costs the run.** A LangChain agent is routinely handed a dozen or more tools at once, and each wrong pick spends an iteration against the `maxIterations` bound LC-111 checks for — so a run can exhaust its step budget and return nothing useful.

```
$ trustabl rules validate .
```

Fire (one `tool(...)` described `"TODO: describe this tool."`, one `"Gets data."`): `LC-021, LC-022, LC-201`
Silent (full description naming when to prefer the neighboring tool): `LC-201`

(LC-201 is the pre-existing missing-AGENTS.md repo rule.)

LC-022 pairs `description_length_lt: 40` with `has_docstring: true` so it doesn't double-report against LC-010 on an empty description, same as CSDK-018.

**Merge note:** this and #66 both append to `langchain/tool_definition.yaml`, so whichever lands second needs a trivial rebase — the rule blocks are independent and the resolution is "keep both." Happy to rebase on request.

### DIFF
--- a/langchain/tool_definition.yaml
+++ b/langchain/tool_definition.yaml
@@ -77,3 +77,63 @@ rules:
       Pass a one-sentence description in the tool() config (or the
       DynamicStructuredTool options) naming what the tool does, the inputs it
       expects, and what it returns. Write it for the model, not a human reader.
+
+  - id: LC-021
+    title: TypeScript LangChain tool description is a placeholder
+    severity: low
+    confidence: 0.85
+    language: typescript
+    applies_to:
+      - langchain_tool
+    scope: tool
+    match:
+      has_description_text:
+        - todo
+        - tbd
+        - fixme
+        - placeholder
+        - no description
+        - does stuff
+    explanation: >
+      The tool sets a description, so it passes LC-010, but the string is a
+      placeholder rather than real content. LangChain.js has no docstring to
+      fall back on — the description field is the entire account of the tool the
+      model sees — so "TODO: describe this tool" leaves the model guessing from
+      the tool name. The Zod schema does not compensate: it constrains the shape
+      of the arguments once the model has chosen this tool, never whether
+      choosing it was right. Mis-selection scales badly here because a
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI, this is a clean one. Checking both rules against `origin/main`: `has_description_text`, `has_docstring` and `description_length_lt` all exist in `internal/rules/schema.go`, `langchain_tool` is a valid `applies_to` at `scope: tool`, and `LC-021` / `LC-022` are free on main and unclaimed by any other open PR. Severity and confidence line up with `LC-001` and `LC-010` next door, and pairing `description_length_lt: 40` with `has_docstring: true` so that an absent description stays `LC-010`'s finding rather than double reporting is exactly right, the same shape as `CSDK-018`. No new predicates, so you are correct that no `schema_version` bump is needed here.

The engine half in trustabl/trustabl#171 mirrors this file byte for byte (both sides land on the same blob hash), so `rules-sync` stays green as long as the two land together.

One thought rather than a blocker: `todo` is matched as a case-insensitive substring, so a genuine tool described "Add an item to the user's todo list." will fire `LC-021`. `CSDK-017` already ships that needle, so I am not asking you to diverge from it here.

Before this ships it needs a rationale doc in `trustabl-rulebook` for the LangChain pack. That part is on us, and we are happy to pair on it. Looks good to me otherwise.

PR #96, opened 2026-08-24
feat(google_adk): add ADK-117, TypeScript FunctionTool writes to the filesystem

bradAGI:
ADK-004 covers the Python path-safety case; the TypeScript half was missing. Mirrors CSDK-012 — the TS half of the Claude SDK pair — including its coarse-signal caveat, stated in the explanation so the finding is honest about itself: it flags *any* filesystem write, not only unnormalized paths, because TS path-normalization analysis isn't wired yet. Confidence 0.5 to match.

**The callback gate doesn't close this, and that's the mitigation an ADK author would reach for.** ADK-102 and ADK-107 check that `before_tool_callback` is *present at all* — a callback that exists still has to inspect the path itself and know which root is allowed. Being wired up isn't the same as containing anything, so "we have a before_tool_callback" isn't an answer to this finding.

Agent composition widens the reach: any branch of the tree that lists this tool can reach it, and a write performed in one branch is visible to every agent that reads the shared session state afterward. That's why the fix puts the check **inside the tool** rather than in a callback — it then holds for every agent in the tree that lists it, instead of only the ones whose callback was configured.

```
$ trustabl rules validate .
```

Fire (`writeFileSync(notePath, body)` in a `new FunctionTool({ ..., execute })` handler): `ADK-117, ADK-201`
Silent (server-derived id, no model-supplied path): `ADK-201`

(ADK-201 is the pre-existing missing-AGENTS.md repo rule.)

Numbered ADK-117 to avoid ADK-111 in the open PR #51 and the IDs used by my other open PRs (#79, #86, #89).

### DIFF
--- a/google_adk/path_safety.yaml
+++ b/google_adk/path_safety.yaml
@@ -32,3 +32,36 @@ rules:
     fix: >
       Resolve the path with `Path(...).resolve()` and assert it sits under an
       allowed root before opening.
+
+  - id: ADK-117
+    title: TypeScript FunctionTool writes to the filesystem
+    severity: low
+    confidence: 0.5
+    language: typescript
+    applies_to:
+      - adk_function_tool
+    scope: tool
+    match:
+      has_write_call: true
+    explanation: >
+      This TypeScript ADK FunctionTool writes to the filesystem. If the path or
+      the contents derive from the tool's arguments, the model chooses both, and
+      a prompt injection carried in retrieved content or an earlier tool result
+      can steer the write at any file the host process can reach. The callback
+      gate the ADK offers does not close this by default: ADK-102 and ADK-107
+      are about before_tool_callback being present at all, and a callback that
+      is present still has to inspect the path itself and know which root is
+      allowed — being wired up is not the same as containing anything. Agent
+      composition widens the reach too, since any branch of the tree that lists
+      this tool can reach it, and a write performed in one branch is visible to
+      every agent that reads the shared session state afterward. (Coarse signal
+      — it flags any filesystem write, not only unnormalized paths, because
+      TypeScript path-normalization analysis is not yet wired. Confirm the path
+      is genuinely model-supplied before acting.)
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI, this is a well built rule. Mirroring `CSDK-012` on `severity: low` / `confidence: 0.5` / `has_write_call: true` is the right calibration, the coarse-signal caveat is stated in the explanation rather than left implicit, and the framing is genuinely ADK's own rather than lifted text: the point that `ADK-102` and `ADK-107` only check that `before_tool_callback` exists is correct, and putting the containment inside the tool is the right prescription. `adk_function_tool` is valid at tool scope, and `has_write_call` resolves on the TypeScript path through the `writes_fs` fact that `tsHandlerFacts` sets, so this is a real signal rather than a Python-only predicate borrowed into a TS rule. Pairing the engine half on a branch of the same name (trustabl/trustabl#174) so `rules-sync` resolves it is exactly right.

One change before this lands. The README documents the ID blocks as `NNN` tool scope, `1NN` agent or subagent scope, `2NN` repo scope, and in `google_adk` that split is currently unbroken: every ADK `1NN` is agent scoped, and the shipped TypeScript tool rules (`ADK-013`, `ADK-015`, `ADK-016`) all sit in the `0NN` block. `ADK-014` is free on `main` and unclaimed by any other open PR, so please renumber to `ADK-014` here and in trustabl/trustabl#174. IDs are permanent once shipped, so this is much cheaper now. Your other tool-scoped ADK PRs have the same issue and I will follow up there. Please also extend the README file map line for `google_adk/path_safety.yaml` to name the new TypeScript rule, the way the `claude_sdk/path_safety.yaml` line already names `CSDK-012`.

Not a request on you: a new rule also needs a rationale doc in `trustabl-rulebook` before it ships, and we can write that or pair with you on it. Happy to take this once the ID is renumbered and the README map line is updated.

PR #97, opened 2026-08-24
feat: add path-containment rules to the four packs that lacked them

bradAGI:

## The gap

Four packs ship a rule for a caller-supplied path reaching filesystem I/O without being resolved and gated:

| pack | rule | severity |
|---|---|---|
| Claude Agent SDK | CSDK-004 | high |
| OpenAI Agents SDK | OAI-006 | high |
| Google ADK | ADK-004 | high |
| MCP | MCP-005 | high |

LangChain, CrewAI, AutoGen and Pydantic AI ship none. This is the **highest-severity** gap in those packs — a path-traversal check simply absent from four of the five listed as weakest coverage. They already have `shell_safety`, `code_execution` and `ssrf`; filesystem containment was the missing member of that set.

Adds **LC-008, CREW-008, AG2-014, PYD-009**, reusing `call_uses_unnormalized_path_param` with the same callee set (`open`, `Path`, `shutil.*`, `os.*`) at the same `high` / `0.7`.

## Why the text is per-SDK

The predicate is shared; what makes the argument *attacker-controlled* is not. Each rule names its own reachability path:

- **LangChain** — the model fills the argument from its context: retrieved documents, tool output, user text.
- **CrewAI** — that context includes other agents' task output, so the path can originate in an agent that summarized an untrusted document.
- **AutoGen** — the argument comes from the shared conversation, which in a group chat *every* participant writes to.
- **Pydantic AI** — worth stating explicitly, because the framework invites the mistake: the argument **is** validated, against its annotation. But `str` and `Path` both accept `../../etc/passwd`. The type constrains what the value looks like, not where it points.

## Verified end to end

Built the engine and scanned real sample projects with `--rules-repo` pointed at this branch:

```
LC-008   [high] read_doc    tools.py:8
CREW-008 [high] read_doc    tools.py:8
AG2-014  [high] read_notes  tools.py:8
PYD-009  [high] read_ledger tools.py:9
```

Each sample also contains a second tool that does `Path(path).resolve()` plus an `is_relative_to(ROOT)` check. None of the rules fired on it — including the per-param behavior the predicate comment documents.

## IDs

`LC-008`, `CREW-008`, `AG2-014`, `PYD-009` — all previously unused.

### DIFF
--- a/autogen/path_safety.yaml
+++ b/autogen/path_safety.yaml
@@ -1,0 +1,42 @@
+policy:
+  id: autogen_path_safety
+  name: AutoGen filesystem path safety
+  category: autogen
+  description: >
+    Rules that catch a model-chosen filesystem path flowing into I/O inside an
+    AutoGen / AG2 tool without containment. Tool arguments are produced by the
+    model from the shared conversation, so a traversal payload reaches real
+    filesystem state if the function does not resolve and gate it.
+
+rules:
+  - id: AG2-014
+    title: AutoGen tool uses a caller-supplied path without containment
+    severity: high
+    confidence: 0.7
+    language: python
+    applies_to:
+      - autogen_tool
+    scope: tool
+    # call_uses_unnormalized_path_param is per-param: a tool with two path
+    # params and one .resolve() correctly fires on the unresolved one.
+    match:
+      call_uses_unnormalized_path_param:
+        callees:
+          - open
+          - Path
+        callee_prefixes:
+          - shutil.
+          - os.
+    explanation: >
+      The tool takes a path-like parameter and hands it to file or directory
+      operations without resolving it or checking it against an allowed root.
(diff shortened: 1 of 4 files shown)


bradAGI:
Closing in favor of the per-rule PRs that supersede this one.

This bundled four path-containment rules; they have since been re-cut individually (AG2-014, PYD-009), and more importantly the IDs here collide with the scheme those PRs use — LC-008 and CREW-008 are assigned to *error handling* there, not path safety. Two rules sharing an ID would be rejected by the loader, so this cannot merge alongside them.

The gap itself is real and still worth closing: LangChain, CrewAI, AutoGen and Pydantic AI ship no path-containment rule while Claude, OpenAI, ADK and MCP each ship one at high severity. The rule bodies, the per-SDK reachability arguments, and the verified fire/silent samples are preserved in the commit history here if they are useful.

PR #98, opened 2026-08-24
feat(langchain): add LC-009, tool network call has no timeout

bradAGI:

## The gap

Every other pack has a no-timeout rule. LangChain does not:

|---|---|---|---|---|
| Claude Agent SDK | CSDK-003 | | AutoGen | AG2-012 |
| OpenAI Agents SDK | OAI-005 | | Pydantic AI | PYD-006 |
| Google ADK | ADK-003 | | Vercel AI | VAI-011 |
| MCP | MCP-004 | | **LangChain** | **— none —** |

`LC-009` uses the same `call_without_kwarg` predicate and the same 20-callee set as PYD-006, at the same `high` / `0.85`.

LangChain has an iteration bound and no time bound. `max_iterations` caps how many steps an agent takes, **not how long one step lasts**, and no LangChain setting caps the duration of a tool call.

So a hung request is not a slow path — it is an unbounded one. The executor blocks inside the step: nothing fails, nothing retries, and the model is never told. The agent simply stops, holding whatever runtime hosts it until the connection eventually dies. That is a worse outcome than the same defect in a plain script, where at least the caller can see it hang.

## Verified end to end

Built the engine and scanned a sample with `--rules-repo` pointed at this branch:

```
LC-009 [high] fetch_status tools.py:6
```

The sample's second tool is the identical request with `timeout=10`; the rule did not fire on it.

### DIFF
--- a/langchain/network.yaml
+++ b/langchain/network.yaml
@@ -1,0 +1,57 @@
+policy:
+  id: langchain_network
+  name: LangChain tool network hygiene
+  category: langchain
+  description: >
+    Network-call hygiene inside LangChain tool functions. A tool that makes a
+    network request without a timeout can hang the agent run indefinitely.
+
+rules:
+  - id: LC-009
+    title: LangChain tool network call has no timeout
+    severity: high
+    confidence: 0.85
+    language: python
+    applies_to:
+      - langchain_tool
+    scope: tool
+    match:
+      call_without_kwarg:
+        callees:
+          - requests.get
+          - requests.post
+          - requests.put
+          - requests.delete
+          - requests.patch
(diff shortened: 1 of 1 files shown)


bradAGI:
Closing as a duplicate.

The no-timeout gap this fills is already covered by the LC-007/LC-015 and CREW-007 network PRs, which reached the same conclusion with different IDs. Keeping both would put two rules on one ID and break the loader.

The LangChain- and CrewAI-specific arguments are preserved in the commit here — that max_iterations/max_iter bounds steps and not time, and that in a sequential CrewAI process a stalled call blocks every task queued behind it.

PR #99, opened 2026-08-24
feat(crewai): add CREW-009, tool network call has no timeout

bradAGI:

## The gap

Every other pack has a no-timeout rule. CrewAI does not:

|---|---|---|---|---|
| Claude Agent SDK | CSDK-003 | | AutoGen | AG2-012 |
| OpenAI Agents SDK | OAI-005 | | Pydantic AI | PYD-006 |
| Google ADK | ADK-003 | | Vercel AI | VAI-011 |
| MCP | MCP-004 | | **CrewAI** | **— none —** |

`CREW-009` uses the same `call_without_kwarg` predicate and the same 20-callee set as PYD-006, at the same `high` / `0.85`.

CrewAI bounds iterations, not time. `max_iter` caps how many steps an agent takes; nothing caps how long one takes, so a hung request is an unbounded path rather than a slow one.

The crew structure makes it worse than the single-agent case. In a sequential process, tasks run in order — so a stall inside one agent's tool call **blocks every task queued behind it**. One unresponsive endpoint stops the whole crew, and none of the downstream tasks produce a result either. The blast radius of a missing `timeout=` is the run, not the call.

## Verified end to end

Built the engine and scanned a sample with `--rules-repo` pointed at this branch:

```
CREW-009 [high] fetch_quote tools.py:6
```

The sample's second tool is the identical request with `timeout=10`; the rule did not fire on it.

### DIFF
--- a/crewai/network.yaml
+++ b/crewai/network.yaml
@@ -1,0 +1,56 @@
+policy:
+  id: crewai_network
+  name: CrewAI tool network hygiene
+  category: crewai
+  description: >
+    Network-call hygiene inside CrewAI tool functions. A tool that makes a
+    network request without a timeout can hang the agent run indefinitely.
+
+rules:
+  - id: CREW-009
+    title: CrewAI tool network call has no timeout
+    severity: high
+    confidence: 0.85
+    language: python
+    applies_to:
+      - crewai_tool
+    scope: tool
+    match:
+      call_without_kwarg:
+        callees:
+          - requests.get
+          - requests.post
+          - requests.put
+          - requests.delete
+          - requests.patch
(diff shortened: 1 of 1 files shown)


bradAGI:
Closing as a duplicate.

The no-timeout gap this fills is already covered by the LC-007/LC-015 and CREW-007 network PRs, which reached the same conclusion with different IDs. Keeping both would put two rules on one ID and break the loader.

The LangChain- and CrewAI-specific arguments are preserved in the commit here — that max_iterations/max_iter bounds steps and not time, and that in a sequential CrewAI process a stalled call blocks every task queued behind it.

PR #102, opened 2026-08-24
feat(vercel_ai): detect privileged tools without approval

CodingWCal:
## Change  Adds VAI-013 for Vercel AI SDK `tool()` and `dynamicTool()` definitions that shell out or execute dynamic code without a tool-level `needsApproval` gate. It fires when the option is missing or literally `false`; `true` and callback gates remain silent.  ## Scope boundary  Unlike OAI-014, VAI-013 deliberately omits `has_write_call`. The Vercel AI pack has no shipped filesystem-write capability rule, so this approval gate currently pairs with VAI-001 (subprocess) and VAI-002 (eval / `new Function`). Revisit filesystem-write approval gating once a Vercel write-capability rule exists.  AI SDK 7 moved approval to call or agent-level `toolApproval`. This tool-scoped rule cannot correlate that setting to an individual tool, so SDK 7 findings require manual review when approval may be configured elsewhere.  The rationale document belongs in the rulebook companion PR, not this YAML-only rules repository.  ## Validation  - [x] Production rule commits: `2bddcde`, `537731a`, `fc649ca`, `d8e176f`, `d6584a8` - [x] Production YAML and engine fixture are line-for-line equivalent - [x] Existing targeted and full Go coverage passed before the final text-only synchronization; no predicate or test behavior changed afterward - [x] Rulebook check adds no new errors; known baseline remains `OAI-112` and `PYD-106`  ## Companion PRs  - Engine: https://github.com/trustabl/trustabl/pull/121 - Rulebook: https://github.com/trustabl/trustabl-rulebook/pull/82  ## Team  - Justin (@jj-javascript): production rule - Calvin: engine integration, scope correction, and release coordination - Karlee (@kperpignant): consulting partner, threat model, and QA review  ## Maintainer decision  VAI-013 is also claimed by rules PR #69 and engine PR #148 for a different description-quality rule. We have not renumbered this security rule; the maintainer can keep the first accepted ID and ask the other pair to rebase and renumber.


jhumel-code:
Thanks @CodingWCal, this is a clean pair. The YAML here is identical to the fixture copy in trustabl/trustabl#121, every predicate it uses (`has_shell_call`, `has_code_exec_call`, `tool_decorator_kwarg_present`, `tool_decorator_kwarg_value`) already exists on `main`, and no `schema_version` bump is needed or present, which is right. Vercel discovery drops unconsumed `tool()` options into `ToolDef.Config`, so a `needsApproval` arrow function lands there as non-literal text and neither branch of the inner `any` matches, exactly as the explanation promises. The silent case for the callback form in the engine PR covers that path rather than passing vacuously.

Two coordination items. VAI-013 is also claimed by trustabl/trustabl-rules#69 and trustabl/trustabl#148 for a different (description quality) rule, so whichever lands first, the other will need to renumber and rebase. And this has to merge with trustabl/trustabl#121 or `rules-sync` goes red.

One design question: VAI-013 can never fire alone, since VAI-001 or VAI-002 already fires at `high` on the same tool. Would `medium` be fairer, so a missing gate does not add a second `high` for one surface?

Good to see trustabl-rulebook#82 alongside this. Happy to pair on the grounding doc before it ships.

sairenchristianbuerano:
Heads up — **we renumbered this rule to `VAI-018`** and pushed the change to your branch, so no action is needed from you.

`VAI-013` had been claimed twice: by this rule (privileged tool with no `needsApproval`) and by the description-quality rule in trustabl/agent-reliability-rules#69, which takes `VAI-013` and `VAI-014` as a contiguous pair. Only one could hold the id, and merging both would have produced a duplicate rule id that the pack validator rejects.

We moved this one because it is a single rule rather than a pair, and because #69 was opened first in the rules repo, which is where ids are allocated. `VAI-018` is the next free id: 001–012 and 016–017 are on `main`, 013/014 go to #69, and 015 is claimed by rules#70.

**Only the id changed.** Title, severity, confidence, match expression, explanation and fix text are all untouched. Both PRs were also rebased onto current `main`.

Verified: all five fire/silent cases pass under the new id, the rules pack and engine fixture are byte-identical, and the validator reports 220 rules valid with both this pair and #69 applied — no duplicate ids.

sairenchristianbuerano (approved):

PR #103, opened 2026-08-24
feat(langchain): add LC-103, agent wires a raw HTTP Requests built-in tool

jiaxinaspenlin-dotcom:
## Summary

Adds LC-103 for LangChain agents that wire `langchain_community`'s `Requests*` built-in tools directly into an agent.

These tools provide outbound HTTP capability where the method is fixed by the class but the destination can be model-controlled, creating exposure to internal services, cloud metadata endpoints, localhost/admin interfaces, and other resources reachable from the agent host.

## What changed

- Added LC-103 to `langchain/agent_safety.yaml`
- Detects:
  - `RequestsGetTool`
  - `RequestsPostTool`
  - `RequestsPutTool`
  - `RequestsPatchTool`
  - `RequestsDeleteTool`
- Uses the existing `agent_uses_hosted_tool_class` predicate
- No new predicate
- No schema version bump

## Rule metadata

- Rule: `LC-103`
- Severity: `medium`
- Confidence: `0.75`
- Language: `python`
- Scope: `agent`

## Validation

`trustabl rules validate ../trustabl-rules`

Result:

`OK: 85 rule pack(s), 207 rule(s) valid under rule schema version 14`

The engine's existing LangChain hosted-tool discovery was also verified to recognize all five `Requests*` classes without analyzer changes.

## Paired changes

- trustabl/trustabl — engine fixture and fire/silent test coverage
- trustabl/trustabl-rulebook#79 — rationale documentation

All coordinated changes use:

`feat/langchain-requests-builtin-tool`

### DIFF
--- a/langchain/agent_safety.yaml
+++ b/langchain/agent_safety.yaml
@@ -83,3 +83,44 @@ rules:
       Pass maxIterations to the AgentExecutor options, sized to the task, and set
       handleParsingErrors so a malformed step is surfaced rather than retried
       indefinitely.
+
+  - id: LC-103
+    title: LangChain agent wires a raw HTTP Requests built-in tool
+    severity: medium
+    confidence: 0.75
+    language: python
+    applies_to:
+      - langchain_agent
+      - langchain_agent_executor
+      - langchain_state_graph
+    scope: agent
+    match:
+      agent_uses_hosted_tool_class:
+        - RequestsGetTool
+        - RequestsPostTool
+        - RequestsPutTool
+        - RequestsPatchTool
+        - RequestsDeleteTool
+    explanation: >
+      This agent's tools list includes a langchain_community Requests built-in
+      (RequestsGetTool / RequestsPostTool / RequestsPutTool / RequestsPatchTool /
+      RequestsDeleteTool). These tools issue an outbound HTTP request to a URL the
+      model supplies; only the method is fixed by the class. Because the
+      destination is model-chosen rather than pinned by the tool, a prompt
+      injection in content the agent has already read can steer the request at
+      whatever the agent host can reach — the cloud metadata endpoint
+      (169.254.169.254), a localhost admin port, or an internal service no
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @jiaxinaspenlin-dotcom. LC-103 fills a real gap: LC-101 covers the REPL and shell built-ins, but the `Requests*` family was uncovered even though the engine's LangChain hosted-tool discovery already recognizes all five classes, so no analyzer change and no schema bump was the right call. `agent_uses_hosted_tool_class` is dispatched at agent scope, the three `applies_to` tokens are all valid for `scope: agent`, and `medium` / `0.75` sits sensibly between LC-101 (`high` / `0.85`) and LC-102 (`low` / `0.6`). The explanation names the concrete consequence (metadata endpoint, localhost admin ports, response bodies re-entering the conversation as untrusted text) and the fix prescribes specific changes, both in LangChain's own vocabulary. The fixture mirror in trustabl/trustabl#175 is identical to this hunk line for line, and the rationale doc is already open as trustabl/trustabl-rulebook#79, which is the part that usually goes missing.

Two small things. The rule is appended after LC-111, so the file is no longer in ID order; moving it under LC-102 reads better. And the common `RequestsToolkit(...).get_tools()` shape does not put the class directly in `tools=[...]`, so this rule will miss that path. That is a discovery gap rather than a bug here, worth a follow-up.

LC-103 is also claimed by trustabl/trustabl-rules#108 and trustabl/trustabl-rules#113, so whichever lands first, the others will need to renumber and rebase.

sairenchristianbuerano:
**Selected for merge.** This pair keeps `LC-103`.

Closing trustabl/agent-reliability-rules#113 (a human-review approval gate, a different rule) and trustabl/agent-reliability-rules#108 (the same rule but matching only `RequestsGetTool`; this PR covers all five Requests classes).

sairenchristianbuerano (approved):

PR #104, opened 2026-08-24
feat(langchain): add LC-024, ambiguous tool name

bradAGI:
Ports the ambiguous-name check to LangChain. Claude SDK (CSDK-007), OpenAI (OAI-007), ADK (ADK-007), and MCP (MCP-003) all ship it; the newer packs had none.

The tool name sits directly beside the description in what the model sees, so **it is half the selection signal** — and a generic verb like `process` or `handle` spends that half on nothing. A LangChain agent is routinely handed a dozen or more tools at once, so a name that does not discriminate competes against every neighbor in the list — and each wrong pick spends an iteration against LC-102's `max_iterations` budget.

```
$ trustabl rules validate .
```

Fire (tool named `process`): includes `LC-024`
Silent (same tool renamed `summarize_invoice`): `LC-024` absent

```go
if tl.Name != "" {
}
```

### DIFF
--- a/langchain/tool_definition.yaml
+++ b/langchain/tool_definition.yaml
@@ -77,3 +77,34 @@ rules:
       Pass a one-sentence description in the tool() config (or the
       DynamicStructuredTool options) naming what the tool does, the inputs it
       expects, and what it returns. Write it for the model, not a human reader.
+
+  - id: LC-024
+    title: Ambiguous LangChain tool name
+    severity: low
+    confidence: 0.9
+    language: python
+    applies_to:
+      - langchain_tool
+    scope: tool
+    match:
+      name_in:
+        - process
+        - handle
+        - run
+        - do
+        - execute
+        - perform
+        - work
+        - go
+        - thing
+        - stuff
+    explanation: >
+      The tool name is a generic verb that says nothing about what the tool
+      acts on. The name sits directly beside the description in what the model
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI, this is a clean port. LC-024 matches CSDK-007 exactly on the `name_in` list, `severity`, and `confidence`, which is what I want for a list that is now shared across five packs, and `langchain_tool` is a valid `applies_to` at `scope: tool` with `name_in` already in the schema, so nothing here needs a `schema_version` bump. The LC-102 cross reference checks out (that is the `max_iterations` rule), and LC-024 is the only free number left in the LC tool range given the other LangChain PRs in flight, so no collision to sort out.

Two follow-ups rather than blockers. LangChain.js tools do carry a name (`buildTSLangChainTool` reads the config object's `name` literal), so a `language: typescript` twin of this rule would work where the Vercel one structurally could not. And yes, please open the engine PR that back-fills `ToolDef.Name` from the Vercel tools-record key.

One nit: the last sentence of `explanation` is a single long unwrapped line, so please wrap it to match the pack. Also, your `rules validate` output reports 211 rules where main plus this one comes to 207, was that run against a tree carrying your other branches?

Before this reaches users it needs a rationale doc in `trustabl-rulebook`. That part is on us, and I am happy to pair on it. I will take this together with trustabl/trustabl#176.

PR #106, opened 2026-08-24
feat(autogen): add AG2-019, ambiguous tool name

bradAGI:
Ports the ambiguous-name check to AutoGen. Claude SDK (CSDK-007), OpenAI (OAI-007), ADK (ADK-007), and MCP (MCP-003) all ship it; the newer packs had none.

The tool name sits directly beside the description in what the model sees, so **it is half the selection signal** — and a generic verb like `process` or `handle` spends that half on nothing. Each wrong pick costs a full round of the conversation — propose, execute, reply — so an ambiguous name spends the AG2-004 `max_round` budget rather than the chat reaching an answer.

```
$ trustabl rules validate .
```

Fire (tool named `process`): includes `AG2-019`
Silent (same tool renamed `summarize_invoice`): `AG2-019` absent

```go
if tl.Name != "" {
}
```

### DIFF
--- a/autogen/tool_definition.yaml
+++ b/autogen/tool_definition.yaml
@@ -54,3 +54,34 @@ rules:
       Annotate every parameter with a concrete type (str, int, list[str], a
       typing.Annotated description, etc.). AutoGen propagates these annotations into
       the schema the model sees, so the model emits correctly-shaped arguments.
+
+  - id: AG2-019
+    title: Ambiguous AutoGen tool name
+    severity: low
+    confidence: 0.9
+    language: python
+    applies_to:
+      - autogen_tool
+    scope: tool
+    match:
+      name_in:
+        - process
+        - handle
+        - run
+        - do
+        - execute
+        - perform
+        - work
+        - go
+        - thing
+        - stuff
+    explanation: >
+      The tool name is a generic verb that says nothing about what the tool
+      acts on. The name sits directly beside the description in what the model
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI, this is a clean port. I checked the `name_in` list against CSDK-007 on `main` and it matches item for item, and `severity: low` with `confidence: 0.9` lines up with AG2-007 in the same file, so nothing here is an outlier. `autogen_tool` is valid at `scope: tool`, `name_in` already exists on `main`, and since no new predicate is involved there is correctly no `schema_version` bump.

What I like most is that the rule is genuinely accurate for AutoGen rather than just plausible. `DiscoverAutoGenTools` honors the `name=` override on both `register_for_llm` and `register_function`, so `ToolDef.Name` holds the name the model actually sees, and a `def process` registered as `name="summarize_invoice"` will not fire.

One nit: the closing sentence of `explanation` is a single long line while the rest of the pack wraps near 75 columns. Please rewrap it.

AG2-019 will also need a rationale block in the `trustabl-rulebook` doc for this pack before it ships. That is on us rather than on you, and we are happy to pair on it.

Your Vercel note checks out too. `ts_vercel_tools.go` never sets `Name` and the test asserts it stays empty, so an engine PR backfilling it from the tools-record key would be welcome as its own change.

This pairs with trustabl/trustabl#178, whose fixture copy is byte identical, so the two should land together.

PR #107, opened 2026-08-24
feat(pydantic_ai): add PYD-013, ambiguous tool name

bradAGI:
Ports the ambiguous-name check to Pydantic AI. Claude SDK (CSDK-007), OpenAI (OAI-007), ADK (ADK-007), and MCP (MCP-003) all ship it; the newer packs had none.

The tool name sits directly beside the description in what the model sees, so **it is half the selection signal** — and a generic verb like `process` or `handle` spends that half on nothing. Pydantic AI derives the tool name from the function name unless one is passed explicitly, so a generically named helper function silently becomes a generically named tool — the defect arrives without anyone deciding to name a tool `process`.

```
$ trustabl rules validate .
```

Fire (tool named `process`): includes `PYD-013`
Silent (same tool renamed `summarize_invoice`): `PYD-013` absent

```go
if tl.Name != "" {
}
```

### DIFF
--- a/pydantic_ai/tool_definition.yaml
+++ b/pydantic_ai/tool_definition.yaml
@@ -62,3 +62,34 @@ rules:
       list[str], a Pydantic model, etc.). Pydantic AI propagates these
       annotations into the schema the model sees, so the model emits
       correctly-shaped arguments and validation stops rejecting them.
+
+  - id: PYD-013
+    title: Ambiguous Pydantic AI tool name
+    severity: low
+    confidence: 0.9
+    language: python
+    applies_to:
+      - pydantic_ai_tool
+    scope: tool
+    match:
+      name_in:
+        - process
+        - handle
+        - run
+        - do
+        - execute
+        - perform
+        - work
+        - go
+        - thing
+        - stuff
+    explanation: >
+      The tool name is a generic verb that says nothing about what the tool
+      acts on. The name sits directly beside the description in what the model
+      sees, so it is half the selection signal, and a name like process or
+      handle spends that half on nothing — the model either calls the tool for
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @bradAGI, this is a clean port. Checking `PYD-013` against `main`: the ID is free, every required field is present, `pydantic_ai_tool` is a valid `applies_to` at `tool` scope, `name_in` already exists in the engine so no `schema_version` bump is needed here, and `low` with `0.9` matches both `CSDK-007` and `PYD-001` in this pack. `PredNameIn` compares the whole name case insensitively, so the list behaves the way your description says it does.

One question before this goes in. Discovery honors an explicit `name=` on the `Tool(...)` factory, but the decorator path in `internal/analysis/discovery.go` builds `ToolDef.Name` straight from the function name and never reads the decorator kwargs. So if `@agent.tool` takes a `name=` override, `@agent.tool(name="summarize_invoice")` over `def process(...)` would fire here even though the name the model actually sees is fine. Decorator kwargs do land in `Config`, so a `not: tool_decorator_kwarg_present: [name]` guard would close it. `CSDK-007` has the same gap, so this is not something you introduced.

Small nit, the tail of the `explanation` runs as one long unwrapped line; please rewrap it to match the rest of the pack. Also worth knowing that trustabl/trustabl-rules#64 appends `PYD-010` and `PYD-011` to the end of this same file, so whichever lands first leaves the other needing a rebase. This will want a rationale doc in `trustabl-rulebook` before it ships, which we can handle or pair on. I have reviewed the engine half, trustabl/trustabl#179, alongside this.

sairenchristianbuerano:
LGTM. This is helpful. Please fix the merge conflicts even in other PR. Thank you very much.

PR #109, opened 2026-08-24
feat(vercel_ai): add VAI-010 mutating tool idempotency rule

joeymussalli:
## Summary
- Add `vercel_ai/idempotency.yaml` with **VAI-010**: TypeScript mutating Vercel AI tools (`create`/`send`/`charge`/…) with no idempotency marker in the tool definition.
- Vercel leaves `ToolDef.Name` empty (model-facing name is the agent's tools-record key), so matching relies on `VarName` via `name_has_prefix` — pairs with the engine PR that falls back to `VarName` when `Name` is empty.

## Stack
1. Engine: VarName fallback + fixture + fire/silent cases
2. **This PR** — rule YAML
3. Rulebook: rationale doc

## Test plan
- [ ] Engine CI green with paired engine branch (or merge engine first)
- [ ] `trustabl rules validate` over the packs
- [ ] Confirm no ID collision with open VAI-013..015 PRs

### DIFF
--- a/vercel_ai/idempotency.yaml
+++ b/vercel_ai/idempotency.yaml
@@ -1,0 +1,56 @@
+policy:
+  id: vercel_ai_idempotency
+  name: Vercel AI SDK mutating-tool idempotency
+  category: vercel_ai
+  description: >
+    Rules that flag mutating Vercel AI SDK tools without an idempotency key.
+    Agents retry tool calls under timeouts and ambiguous failures; without a
+    key the same side-effecting action can fire more than once. Vercel tools
+    leave ToolDef.Name empty (the model-facing name is the agent's tools-record
+    key), so matching uses the binding identifier (VarName) via name_has_prefix.
+
+rules:
+  - id: VAI-010
+    title: TypeScript mutating Vercel AI tool has no idempotency key
+    severity: medium
+    confidence: 0.5
+    language: typescript
+    applies_to:
+      - vercel_ai_tool
+    scope: tool
+    match:
+      all:
+        - name_has_prefix:
+            - create
+            - send
+            - delete
+            - post
+            - update
+            - refund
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @joeymussalli, this is well put together. Modeling it on `OAI-019` rather than widening that rule's `applies_to` is the right call for our SDK-scoped model, the `explanation` and `fix` text is genuinely Vercel-specific rather than copied across, `vercel_ai_tool` is a valid tool-scope `applies_to`, severity and confidence line up with `OAI-019`, and `VAI-010` is a free slot in the series (main has `VAI-001` through `VAI-009`, plus `VAI-011` and `VAI-012`), so there is no collision with the contested `VAI-013` PRs.

One thing I want fixed first. `has_body_text` is case-sensitive (plain `strings.Contains`, unlike `has_description_text`), and your `fix` text tells authors to send a Stripe `Idempotency-Key` header, a REST `X-Request-ID`, or a GraphQL `clientMutationId`. None of those spellings are in the needle list, so a tool that follows your own advice can still fire. Please add `Idempot`, `Request-ID`, and `clientMutationId`. `OAI-019` has the same gap and I will bring it in line separately.

On ordering, this needs trustabl/trustabl#182 to land first or alongside: the fixture mirror lives there (byte-identical to this file, which is what I want to see), and without that PR's `VarName` fallback `name_has_prefix` reads the empty `ToolDef.Name` and this rule loads clean but never fires. #182 also carries `VAI-016`, whose production half is trustabl/trustabl-rules#112, so all three want to move together to keep rules-sync green. A rulebook rationale doc is also needed before this ships, but that is not on you, we can handle it or pair on it.

Happy to take this once the needle list is updated.

sairenchristianbuerano (approved):

PR #110, opened 2026-08-24
feat(langchain): add human approval prerequisite rule

dezsow:
Adds LC-112, a LangChain v1 create_agent rule for privileged execution tools (ShellTool, PythonREPLTool, PythonAstREPLTool) that lack the structural prerequisites for a resumable human-approval boundary.
The rule reuses existing agent predicates—no schema or evaluator changes—and checks for non-empty middleware plus a checkpointer when privileged execution is exposed. Companion changes in trustabl mirror the fixture and add regression coverage; the rulebook PR documents the threat model, calibration, and limitations.

### DIFF
--- a/langchain/agent_safety.yaml
+++ b/langchain/agent_safety.yaml
@@ -83,3 +83,37 @@ rules:
       Pass maxIterations to the AgentExecutor options, sized to the task, and set
       handleParsingErrors so a malformed step is surfaced rather than retried
       indefinitely.
+
+  - id: LC-112
+    title: LangChain privileged create_agent lacks human-approval prerequisites
+    severity: high
+    confidence: 0.7
+    language: python
+    applies_to:
+      - langchain_agent
+    scope: agent
+    match:
+      all:
+        - agent_class:
+            - CreateAgent
+        - agent_uses_hosted_tool_class:
+            - PythonREPLTool
+            - PythonAstREPLTool
+            - ShellTool
+        - any:
+            - agent_kwarg_list_empty:
+                - middleware
+            - agent_kwarg_missing:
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks for this, @dezsow. The rule is schema-clean: every required field is set, `langchain_agent` is a valid `applies_to` for `agent` scope, and all four predicates already exist on `main`, so no `schema_version` bump is needed. `agent_kwarg_list_empty` treats an absent kwarg as empty, so the `any` branch reads as "middleware not configured or checkpointer not configured", which is what the title claims. The explanation and fix are framed for LangChain rather than borrowed from another SDK, and I appreciate that the fix says out loud that this only verifies structural prerequisites. The fixture mirror in trustabl/trustabl#181 matches this file exactly.

One question before merging. LC-101 in this same pack already fires at `high` / 0.85 on exactly those three tool classes for `langchain_agent`, so every LC-112 hit is a second high on the same call site, and LC-101's fix already prescribes the human-in-the-loop gate. Would `medium` be fairer here? Worth capturing as a limitation too: on LangGraph Platform the checkpointer comes from the deployment and passing one is discouraged, so a correctly gated agent there still trips that branch.

This needs to land with trustabl/trustabl#181, which is missing its `policies_test.go` coverage entries, so I cannot take it alone yet. LC-112 also needs a rationale doc in `trustabl-rulebook`; that part is ours and we are glad to pair on it.

Happy to merge once the severity question is settled and the engine side is green.

sairenchristianbuerano (changes requested):
Thank you for the contribution, this is very helpful. Please fix the merge conflict.

PR #111, opened 2026-08-24
feat(pydantic): add PYD-104 force_download rule (schema 15)

Paramjeet-singh-neu:
## Summary
- Add **PYD-104** (medium / 0.8 / agent): `FileUrl`-family or `WebFetchTool` with `force_download=True` or `"allow-local"`.
- Bump pack `schema_version` 14 → **15** for the new `agent_file_url_force_download` predicate (older binaries skip this rule and keep the rest).
- Fixture in the engine repo is identical. Companion engine PR: https://github.com/trustabl/trustabl/pull/185

## Test plan
- [x] YAML matches `trustabl` `testdata/rules-fixture/pydantic_ai/agent_safety.yaml`
- [ ] Reviewer: confirm PYD-104 sits between PYD-103 and PYD-105 with no ID collision

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -11,4 +11,4 @@
 #
 # This file is metadata, not a rule: the engine's loader skips manifest.yaml
 # when walking the pack for policy files.
-schema_version: 14
+schema_version: 15
--- a/pydantic_ai/agent_safety.yaml
+++ b/pydantic_ai/agent_safety.yaml
@@ -5,8 +5,8 @@ policy:
   description: >
     Agent-scope safety rules for Pydantic AI agents (the Agent(...) constructor).
     These flag missing output validation, model-driven code execution and URL
-    fetching wired as native tools, and a retry strategy that re-runs an
-    already-executed tool.
+    fetching wired as native tools, a FileUrl force_download that fetches on the
+    agent host, and a retry strategy that re-runs an already-executed tool.
 
 rules:
   - id: PYD-101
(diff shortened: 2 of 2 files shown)


jhumel-code:
Thanks @Paramjeet-singh-neu. PYD-104 is unclaimed on main and no other open PR takes that ID, so it slots between PYD-103 and PYD-105 cleanly. The schema side checks out: every required field is present, `pydantic_ai_agent` is valid at agent scope, and the `value: "True"` / `"allow-local"` form matches how `openai_sdk/approvals.yaml` already uses `agent_hosted_tool_kwarg_value` (the evaluator quote strips string literals, so the `allow-local` arm compares correctly). The YAML is byte identical to the fixture in trustabl/trustabl#185, which adds the `agent_file_url_force_download` predicate and moves `SupportedSchemaVersion` to 15, so nothing here loads against a predicate the engine lacks. The explanation and fix are framed for Pydantic AI and name the concrete consequence rather than gesturing at it.

One change I want. Confidence 0.8 sits above PYD-103's 0.75, but PYD-103 reads a kwarg straight off the agent constructor, while the FileUrl arm here attributes a file level `force_download` to every Pydantic agent in that file. The looser attribution should carry the lower number, so please drop PYD-104 to 0.7 and mirror it in the engine fixture. This also cannot merge ahead of trustabl/trustabl#185, or `rules-sync` goes red and older binaries skip the rule.

PYD-104 will need a rationale doc in `trustabl-rulebook` before it ships. That part is on us and we are happy to pair on it.

Happy to take this once the confidence is adjusted.

sairenchristianbuerano:
Thanks @jhumel-code  — confidence is now 0.7 in both places, and I've rebased both PRs onto current main.

Confidence. Your reasoning holds against the neighbours: PYD-103 reads a kwarg straight off the agent constructor and sits at 0.75, PYD-105 sits at 0.7. The FileUrl arm here attributes a file-level force_download to every Pydantic agent in the file, so 0.7 is the right shelf. Changed in this PR and mirrored in trustabl/agent-reliability-analyzer#185; the two blocks are byte-identical again.

One thing in your review has gone stale — worth flagging before anyone merges. Both PRs were written against pack schema 14 and bumped to 15. main has since moved to 16 on both sides (engine SupportedSchemaVersion = 16, rules manifest.yaml: schema_version: 16). Applying the PRs as written would have downgraded both repos and made every schema-16 rule unloadable.

So the manifest bump is dropped entirely — 16 already covers this rule, and the rules PR is now a single file, +45/−2. The engine PR keeps SupportedSchemaVersion = 16.

Merge order still stands, for the reason you gave. trustabl/agent-reliability-analyzer#185 first, this one second, or rules-sync goes red and older binaries skip the rule.

Confirmed against main: PYD-104 is unclaimed, no other open PR takes the ID, and it slots between PYD-103 and PYD-105.

Commit subject amended from "(schema 15)" since that's no longer what it does.

Rulebook rationale doc still outstanding on your side — happy to pair.

sairenchristianbuerano (approved):

PR #112, opened 2026-08-24
feat(vercel_ai): add VAI-016 ambiguous tool binding name

joeymussalli:
## Summary
- Add **VAI-016** to `vercel_ai/tool_definition.yaml`: ambiguous binding names (`process`/`handle`/`run`/…) via `name_in`.
- Matches `VarName` because Vercel leaves `ToolDef.Name` empty — needs engine VarName fallback (trustabl#182).

## Why
Other SDKs already have ambiguous-name rules (CSDK-007, OAI-007, MCP-003; open PRs for LC/CREW/AG2/PYD). Vercel was the gap; the rulebook explicitly noted it.

## Stack
1. Engine trustabl#182 (VarName + fixtures/tests)
2. **This PR**
3. Rulebook docs PR

## Test plan
- [ ] Engine CI green with #182
- [ ] `trustabl rules validate`

### DIFF
--- a/vercel_ai/tool_definition.yaml
+++ b/vercel_ai/tool_definition.yaml
@@ -58,3 +58,36 @@ rules:
       field per argument (z.object({ city: z.string(), units: z.enum([...]) })).
       Reserve dynamicTool for the rare case where the input genuinely cannot be
       typed, and even then validate the shape inside execute() before using it.
+
+  - id: VAI-016
+    title: Ambiguous Vercel AI tool binding name
+    severity: low
+    confidence: 0.85
+    language: typescript
+    applies_to:
+      - vercel_ai_tool
+    scope: tool
+    match:
+      name_in:
+        - process
+        - handle
+        - run
+        - do
+        - execute
+        - perform
+        - work
+        - go
+        - thing
+        - stuff
+    explanation: >
+      A Vercel AI SDK tool whose binding identifier is a generic name like
+      `process`, `handle`, or `run` gives the model no signal about intent once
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks for this @joeymussalli. The rule is well formed: all required fields are present, `vercel_ai_tool` is a valid tool-scope token, `name_in` already exists on `main`, and low/0.85 sits comfortably beside VAI-004 and the CSDK-007 analog it mirrors. Correct call not to touch `manifest.yaml` as well, since nothing here adds a predicate.

One blocking item: the file loses its trailing newline. `scripts/check-rules-sync.sh` diffs the pack against `testdata/rules-fixture/` with only carriage returns stripped, and your fixture copy in trustabl/trustabl#182 keeps its final newline, so rules-sync will fail with a content drift on `vercel_ai/tool_definition.yaml`. Restore it and the two copies match exactly.

Ordering matters here too. On `main`, `PredNameIn` reads `ToolDef.Name`, which the Vercel discovery leaves empty, so this rule is a silent no-op until trustabl/trustabl#182 lands.

One accuracy point on the text: `directAssignmentName` only captures the `const x = tool(...)` shape, so the idiomatic `tools: { process: tool({...}) }` form yields no `VarName` and the rule stays quiet there. I would narrow the explanation to the binding form rather than the tools record.

This will also need a rationale doc in the rulebook before it ships. Not something to put on you, we can handle it or pair with you on it.

Happy to take this once the trailing newline is back and the explanation is narrowed.

sairenchristianbuerano (approved):

PR #114, opened 2026-08-24
feat(vercel_ai): add VAI-017 tool stdout diagnostics rule

joeymussalli:
## Summary
- Add `vercel_ai/observability.yaml` with **VAI-017**: Vercel AI `execute()` writes diagnostics to stdout (`has_print_call: true`).
- Ports the OAI-010 observability pattern to the Vercel pack. stderr remediations (`console.error`, etc.) do not fire once the engine `prints_stdout` fact is present.

## Depends on
- Engine: https://github.com/trustabl/trustabl/pull/187 (overlaps #180 on the predicate/facts hunk)
- Rulebook: https://github.com/trustabl/trustabl-rulebook/pull/102

## Test plan
- [ ] Pack validates against schema
- [ ] Engine fire/silent cases for VAI-017 pass once #187 (or #180 + fixture) is available

### DIFF
--- a/vercel_ai/observability.yaml
+++ b/vercel_ai/observability.yaml
@@ -1,0 +1,37 @@
+policy:
+  id: vercel_ai_observability
+  name: Vercel AI SDK tool observability hygiene
+  category: vercel_ai
+  description: >
+    Rules covering how a Vercel AI SDK tool emits diagnostics. Tool execute()
+    bodies that write to stdout are invisible to the model and can corrupt
+    stdio-based transports (MCP) that share the channel.
+
+rules:
+  - id: VAI-017
+    title: Vercel AI tool execute() writes diagnostics to stdout
+    severity: low
+    confidence: 0.65
+    language: typescript
+    applies_to:
+      - vercel_ai_tool
+    scope: tool
+    match:
+      has_print_call: true
+    explanation: >
+      This Vercel AI SDK tool's execute() handler writes to the process's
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @joeymussalli. Schema-wise this is correct: `vercel_ai_tool` is valid at `tool` scope, `has_print_call` already exists in the engine, `low` with confidence 0.65 matches OAI-010 exactly, and VAI-017 is not claimed by any other open PR, so there is no renumbering risk here. The YAML matches the fixture in trustabl/trustabl#187 byte for byte, and thanks for opening the rulebook PR alongside rather than leaving the grounding to us.

One change before this ships. The explanation is ported from OAI-010, and both consequences it names are a stretch for this SDK. A Vercel AI tool usually runs in a Next.js route on a platform that captures stdout as its log stream, so the "silently disappears" claim mostly does not hold there, and `ai` package tools are not served over MCP stdio (the SDK consumes MCP as a client). Please lead instead with what does hold: `execute()` receives model-controlled arguments, so a stdout write puts untrusted input, and whatever the tool read from it, into shared logs, and none of it is addressable by the model. ADK-009 is a good in-repo model for that framing.

One note for me rather than for you: on a binary older than trustabl/trustabl#187 this rule loads fine and silently never fires, so I need to decide whether that warrants a `schema_version` bump.

Happy to take this once the explanation is reframed and the fixture is updated to match.

sairenchristianbuerano (approved):

PR #115, opened 2026-08-24
feat(langchain): detect Requests calls without timeout

pjsk02:
## Summary

Adds `LC-007` to detect Python `requests` calls without a finite timeout inside LangChain tools.

LangChain Python currently lacks equivalent timeout coverage while similar network-timeout rules exist for other supported SDKs. This adds that coverage using Trustabl''s existing `call_without_kwarg` predicate without requiring engine or schema changes.

## Detection

`LC-007` fires for supported `requests` calls inside LangChain tools when:

- `timeout` is omitted
- `timeout=None`

It remains silent when an explicit finite timeout is supplied.

The rule intentionally does **not** flag `httpx` merely because a call omits an explicit `timeout`, since HTTPX provides default timeout behavior.

## Scope

This contribution is intentionally limited to Python Requests call-site detection.

Known limitations include wrapper functions, dynamically configured client defaults, positional timeout handling, and HTTP libraries outside the current Requests-specific scope.

## Validation

- Existing `call_without_kwarg` predicate reused
- No production engine changes
- No schema changes
- No manifest changes
- Canonical rule mirrored byte-for-byte in the companion engine fixture
- `git diff --check` passes

## Companion PRs

- Engine fixture/tests: https://github.com/trustabl/trustabl/pull/191
- Rulebook rationale: https://github.com/trustabl/trustabl-rulebook/pull/107

### DIFF
--- a/langchain/network.yaml
+++ b/langchain/network.yaml
@@ -1,0 +1,43 @@
+policy:
+  id: langchain_network
+  name: LangChain tool network hygiene
+  category: langchain
+  description: >
+    Network-call hygiene inside LangChain tool functions. A tool that makes a
+    Requests HTTP call without a finite timeout can remain blocked when an
+    upstream service fails to respond, stalling the agent tool invocation.
+
+rules:
+  - id: LC-007
+    title: LangChain tool network call has no timeout
+    severity: high
+    confidence: 0.85
+    language: python
+    applies_to:
+      - langchain_tool
+    scope: tool
+    match:
+      call_without_kwarg:
+        callees:
+          - requests.get
+          - requests.post
+          - requests.put
+          - requests.delete
+          - requests.patch
+          - requests.head
+          - requests.request
+          - requests.Session.get
+          - requests.Session.post
+        missing: timeout
+    explanation: >
+      This LangChain tool performs a Requests HTTP call without a finite
+      timeout, so it can remain blocked when an upstream service or network
+      fails to respond. The tool invocation cannot complete while the request
+      waits, which consumes worker and runtime resources and prevents the
+      agent step from finishing. A kwarg present with literal None counts as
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @pjsk02, this is careful work. The rule is schema correct with every required field set, `scope: tool` paired with a valid `langchain_tool`, and `severity: high` / `confidence: 0.85` matching every sibling network rule on main. It reuses the existing `call_without_kwarg` predicate rather than inventing one, so leaving `manifest.yaml` untouched is the right call. I diffed the YAML against the fixture copy in trustabl/trustabl#191 and the two are byte identical, which is what the sync job needs.

One change before merge. The callee list is requests only, while the six existing `network.yaml` packs also cover httpx, `urllib.request.urlopen`, bare `urlopen` and the aiohttp ClientSession pair. Your httpx reasoning is fair, since httpx does default to five seconds, but it does not hold for `urllib.request.urlopen`, which blocks forever by default, so a LangChain tool calling `urlopen(url)` slips straight through. Please align the list with the sibling packs, at minimum urllib and aiohttp. Dropping httpx should be one decision across all the packs rather than LangChain alone. Small nit while you are in there: the sentence about a kwarg with literal `None` counting as missing is predicate behavior rather than a user-facing consequence, so I would drop it from `explanation`.

Two coordination notes. `LC-007` is also claimed by trustabl/trustabl-rules#52 and trustabl/trustabl#127, which add the same ID in the same `langchain/network.yaml` path, so whichever lands first the other will need to renumber and rebase. The rule also needs a rulebook rationale doc before it ships; you already opened one, and we can pair on that part if the front matter needs adjusting.

Happy to take this once the callee list is aligned.

sairenchristianbuerano (approved):

PR #116, opened 2026-08-24
feat(claude_skill): add CSKILL-087 placeholder description rule

kureen-cyber:
## Summary
- Adds **CSKILL-087** (low): skill `description:` is a placeholder stub (`TODO` / `TBD` / `FIXME` / `placeholder` / `no description` / `does stuff`).
- Complements CSKILL-070, which only flags an *empty* description. Claude Code still loads a stub into context and uses it as the model-invocation signal, so `"TODO: describe this skill"` is functionally the same as no description.
- Reuses the existing `skill_description_has_text` predicate and the same needle list as CSDK-017. No `schema_version` bump.

Companion PRs (same branch name `feat/cskill-087-placeholder-description`):
- Engine: https://github.com/trustabl/trustabl/pull/192
- Rulebook: https://github.com/trustabl/trustabl-rulebook/pull/109

## Test plan
- [ ] `trustabl rules validate .` on this pack
- [ ] Engine `TestPolicyRules_AllRulesCovered` includes CSKILL-087 fire + silent
- [ ] Fire: `description: "TODO: describe this skill."`
- [ ] Silent: a real description with no placeholder needles

### DIFF
--- a/claude_skill/skill_quality_text.yaml
+++ b/claude_skill/skill_quality_text.yaml
@@ -6,7 +6,8 @@ policy:
     Quality and data-governance signals in Claude Code Agent Skills (SKILL.md)
     detected via free-text matching over the skill's name, description, and
     body — claimed crypto/security purpose, sensitive-data handling, missing
-    error-handling or purpose language, and broad-access or retention phrasing.
+    error-handling or purpose language, placeholder descriptions, and
+    broad-access or retention phrasing.
     Skill rules audit SKILL.md frontmatter + body and are language-agnostic
     (markdown), so they carry no language: field.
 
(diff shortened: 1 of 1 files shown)


jhumel-code:
Thanks @kureen-cyber, this is a clean rule. `CSKILL-087` is the next free ID, the schema side is right (`scope: skill` with `applies_to: claude_skill`, and correctly no `language:` field for a markdown surface), `skill_description_has_text` already exists on `main` so no `schema_version` bump is needed, and `low` / `0.85` matches both the CSDK-017 tool-side analogue and CSKILL-070 next door. The CSKILL-070 gap you describe is real, and the explanation and fix name the concrete consequence and the concrete change.

Two non-blocking notes. That predicate is an unanchored, case-insensitive substring match, so `todo` and `placeholder` will also fire on honest descriptions like "Extracts TODO comments and files them as issues". CSDK-017 carries the same list, so I will not hold this to a different standard, but skill descriptions are longer trigger prose than tool descriptions, so the collision surface is wider. Tightening the needles or dropping confidence to 0.7 would both be welcome. Also worth knowing a stub description will co-fire CSKILL-085, so it reports twice at low.

This still needs a rulebook rationale doc before it ships; the description points at trustabl/trustabl-rulebook#109, and we can pair on that if the front-matter gate fights you. I have the engine half at trustabl/trustabl#192 and its fixture copy is byte identical to this pack, so I will land the two together, this one first.

sairenchristianbuerano (approved):

